PR #302, opened 2022-06-13
feat(CLI): Add go API to support LSP

gaocegege:
This PR is to add golang API, which will be used in the LSP.


gaocegege (commented):
I moved the envd/api/config.py to envd/api/config/__init__.py, not sure if it breaks API doc gen, please have a look @VoVAllen

VoVAllen (approved):
LGTM

PR #321, opened 2022-06-14
fix(release): Change docker user

gaocegege:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,8 +26,8 @@ jobs:
         uses: docker/login-action@v2
         with:
           registry: ghcr.io
-          username: ${{ secrets.CR_USER }}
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -26,7 +26,7 @@ jobs:
         uses: docker/login-action@v1
         with:
           registry: ghcr.io
-          username: ${{ github.repository_owner }}
+          username: ${{ sccrets.DOCKER_USER }}
```
Xuanwo: How about using `CR_USER` and `CR_PAT` so we won't be confused?

For the better `GHCR_USERNAME` & `GHCR_TOKEN`.

gaocegege: SGTM

Xuanwo (approved):

terrytangyuan (approved):

PR #328, opened 2022-06-14
fix: Disable unit test in macOS

gaocegege:
https://github.community/t/why-is-docker-not-installed-on-macos/17017/7

Docker cannot be installed in GitHub macOS runners. Thus we disable the unit test for macOS. We just build the binary.


rudeigerc (approved):
LGTM

Issue #246, opened 2022-06-08
feat(CLI): envd up a GPU image without GPUs


gaocegege:
## Description

Now we consider the user wants a GPU if `cuda` is called in build.envd. But we may need to support `envd up --no-gpu`.

kenwoodjw:
mean ignore cuda command？

gaocegege:
No, users build.envd may look like:

```
def build():
  cuda(xxx)
```

When the user run `envd up`, we will use nvidia-container-runtime to launch a GPU container. 

This feature is to launch a CPU container with the same build.envd when user run `envd up --no-gpu`

gaocegege:
LGTM.

PR #338, opened 2022-06-14
Example: add cpu example

VoVAllen:
Add cpu example & use it as default

### DIFF
--- a/examples/mnist/README.md
+++ b/examples/mnist/README.md
@@ -1,9 +1,22 @@
 # MNIST Example
 
-## Quick start
+## Quick start (CPU)
 
 ```bash
 $ envd up
 ```
+
+By default, `envd up` will use the `build.envd` 
+
+Then you can open jupyter notebook at [`http://localhost:8888`](http://localhost:8888), or open vscode remote to attach to the container.
+
+
+## Quick start (GPU)
+
+```bash
+$ envd up -f build_gpu.envd
+```
+
+Also you can use `-f` option to specify the file to build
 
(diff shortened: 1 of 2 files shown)


VoVAllen:
Later we might want to enable the grammar of `envd up -f buid.envd:build_gpu` for build GPU, and we'll have one envd file looks like 
```
def build():
    install.vscode_extensions([
        "ms-python.python",
    ])

    install.python_packages([
        "tensorflow",
        "numpy",
    ])
    install.cuda(version="11.6", cudnn="8")
    shell("zsh")
    config.jupyter(password="", port=8888)

def build_gpu():
    build()    
    install.cuda(version="11.6", cudnn="8")
```

terrytangyuan:
> Later we might want to enable the grammar of `envd up -f :build_gpu` to build with cuda, and we'll have one single envd file looks like

Relevant issue https://github.com/tensorchord/envd/issues/246. `--no-gpu` seems handy but may not be generic for other use cases. WDYT @gaocegege @VoVAllen?

gaocegege:
@VoVAllen I think the gpu example code in TF is only run on GPUs. Could you please verify if it can be run on CPUs?

gaocegege:
> Relevant issue https://github.com/tensorchord/envd/issues/246. --no-gpu seems handy but may not be generic for other use cases. WDYT @gaocegege @VoVAllen?

Yep, I think we can use a flag here

VoVAllen:
I tested it works with cpu. Tensorflow can automatically switch between CPU and GPU since it doesn't have explicit device placement

gaocegege:
LGTM. 

Please rebase.

VoVAllen:
rebased to master

gaocegege:

Issue #382, opened 2022-06-16
bug:  cannot assign requested address


junnplus:
## Description
```
[N] ❯ envd up
[+] ⌚ parse build.envd and download/cache dependencies 0.0s ✅ (finished)
 => 💽 (cached) download oh-my-zsh                                                                                                                                                                                           0.0s
[+] 🐋 build envd environment 2.9s (20/20) ✅ (finished)
 => docker-image://docker.io/tensorchord/python:3.8-ubuntu20.04                                                                                                                                                             2.1s
 => => resolve docker.io/tensorchord/python:3.8-ubuntu20.04                                                                                                                                                                 2.1s
 => local://cache-dir                                                                                                                                                                                                       0.1s
 => => transferring cache-dir: 70.89kB                                                                                                                                                                                      0.1s
 => 💽 (cached) [internal] create user group envd                                                                                                                                                                            0.0s
 => 💽 (cached) [internal] create user envd                                                                                                                                                                                  0.0s
 => 💽 (cached) [internal] add user envd to sudoers                                                                                                                                                                          0.0s
 => 💽 (cached) [internal] configure user permissions                                                                                                                                                                        0.0s
 => 💽 (cached) [internal] configure user permissions                                                                                                                                                                        0.0s
 => 💽 (cached) install system packages                                                                                                                                                                                      0.0s
 => 💽 (cached) copy /oh-my-zsh /home/envd/.oh-my-zsh                                                                                                                                                                        0.0s
 => 💽 (cached) mkfile /home/envd/install.sh                                                                                                                                                                                 0.0s
 => 💽 (cached) install oh-my-zsh                                                                                                                                                                                            0.0s
 => 💽 (cached) mkfile /home/envd/.zshrc                                                                                                                                                                                     0.0s
 => 💽 (cached) install conda packages                                                                                                                                                                                       0.0s
 => 💽 (cached) install ssh keys                                                                                                                                                                                             0.0s
 => 💽 (cached) install ssh keys                                                                                                                                                                                             0.0s
 => 💽 (cached) [internal] settings pip cache mount permissions                                                                                                                                                              0.0s
 => 💽 (cached) pip install numpy                                                                                                                                                                                            0.0s
 => 💽 (cached) install PyPI packages                                                                                                                                                                                        0.0s
 => 💽 (cached) merging all components into one                                                                                                                                                                              0.0s
 => exporting to oci image format                                                                                                                                                                                           0.8s
[...]

PR #386, opened 2022-06-16
fix: cannot assign requested address

junnplus:
Signed-off-by: Ye Sijun <junnplus@gmail.com>

Fixes: https://github.com/tensorchord/envd/issues/382

I tested it on my local (archlinux) and it works, so does the ubuntu VM.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,7 +24,6 @@ require (
 	github.com/onsi/gomega v1.19.0
 	github.com/opencontainers/go-digest v1.0.0
 	github.com/opencontainers/image-spec v1.0.3-0.20211202183452-c5a74bcca799
-	github.com/phayes/freeport v0.0.0-20220201140144-74d24b5ae9f5
 	github.com/pkg/sftp v1.13.5
 	github.com/sirupsen/logrus v1.8.1
 	github.com/spf13/viper v1.4.0
--- a/go.sum
+++ b/go.sum
@@ -454,8 +454,6 @@ github.com/opentracing/opentracing-go v1.1.0/go.mod h1:UkNAQd3GIcIGf0SeVgPpRdFSt
 github.com/pelletier/go-toml v1.2.0/go.mod h1:5z9KED0ma1S8pY6P1sdut58dfprrGBbd/94hg7ilaic=
 github.com/pelletier/go-toml v1.9.4 h1:tjENF6MfZAg8e4ZmZTeWaWiT2vXtsoO6+iuOjFhECwM=
 github.com/pelletier/go-toml v1.9.4/go.mod h1:u1nR/EPcESfeI/szUZKdtJ0xRNbUoANCkoOuaOx1Y+c=
-github.com/phayes/freeport v0.0.0-20220201140144-74d24b5ae9f5 h1:Ii+DKncOVM8Cu1Hc+ETb5K+23HdAMvESYE3ZJ5b5cMI=
-github.com/phayes/freeport v0.0.0-20220201140144-74d24b5ae9f5/go.mod h1:iIss55rKnNBTvrwdmkUpLnDpZoAHvWaiq5+iMmen4AE=
 github.com/pingcap/errors v0.11.4 h1:lFuQV/oaUMGcD2tqt+01ROSmJs75VG1ToEOkZIZ4nE4=
 github.com/pingcap/errors v0.11.4/go.mod h1:Oi8TUi2kEtXXLMJk9l1cGmz20kV3TaQ0usTwv5KuLY8=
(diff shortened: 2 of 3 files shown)


gaocegege:

The package freeport source code is this:

```go
package freeport

import (
	"net"
)

// GetFreePort asks the kernel for a free open port that is ready to use.
func GetFreePort() (int, error) {
	addr, err := net.ResolveTCPAddr("tcp", "localhost:0")
	if err != nil {
		return 0, err
	}

	l, err := net.ListenTCP("tcp", addr)
	if err != nil {
		return 0, err
	}
	defer l.Close()
	return l.Addr().(*net.TCPAddr).Port, nil
}

// GetPort is deprecated, use GetFreePort instead
// Ask the kernel for a free open port that is ready to use
func GetPort() int {
	port, err := GetFreePort()
	if err != nil {
		panic(err)
	}
	return port
}

// GetFreePort asks the kernel for free open ports that are ready to use.
func GetFreePorts(count int) ([]int, error) {
	var ports []int
	for i := 0; i < count; i++ {
		addr, err := net.ResolveTCPAddr("tcp", "localhost:0")
		if err != nil {
			return nil, err
		}

		l, err := net.ListenTCP("tcp", addr)
		if err != nil {
			return nil, err
		}
		defer l.Close()
		ports = append(ports, l.Addr().(*net.TCPAddr).Port)
	}
	return ports, nil
}
```

gaocegege:
I cannot figure out why it fails. Let me double-check it.

Xuanwo:
> "localhost:0" won't work on my Archlinux, but both ":0" and "127.0.0.0:0" work.

Wired.

Can you give your output of `drill localhost` and `cat /etc/hosts`?

junnplus:
> Can you give your output of `drill localhost` and `cat /etc/hosts`?

```
[N] ❯ drill localhost
;; ->>HEADER<<- opcode: QUERY, rcode: NOERROR, id: 24983
;; flags: qr ; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 0
;; QUESTION SECTION:
;; localhost.   IN      A

;; ANSWER SECTION:
localhost.      30      IN      A       198.18.1.242

;; AUTHORITY SECTION:

;; ADDITIONAL SECTION:

;; Query time: 6 msec
;; SERVER: 198.18.0.2
;; WHEN: Thu Jun 16 20:22:42 2022
;; MSG SIZE  rcvd: 52

[N] ❯ cat /etc/hosts
   1 # Static table lookup for hostnames.
   2 # See hosts(5) for details.
```

pkg/util/netutil/netutil.go
```diff
@@ -14,14 +14,13 @@
...
 package netutil
 
-import (
-	"github.com/phayes/freeport"
-)
+import "net"
 
 func GetFreePort() (int, error) {
-	port, err := freeport.GetFreePort()
+	l, err := net.Listen("tcp", ":0")
```
Xuanwo, line 20: Listen `:0` could make this port open to the public.

```suggestion
	l, err := net.Listen("tcp", "127.0.0.1:0")
```

Xuanwo: BTW, why not use `net.Listen("tcp", ":0")` directly? We can pass a `Listener` to ssh client which is more common in golang.

@gaocegege

Xuanwo:
> ;; ANSWER SECTION:
> localhost.      30      IN      A       198.18.1.242

It's so wired that `localhost` point to IP like `198.18.1.242` instead of `127.0.0.1`.

pkg/util/netutil/netutil.go
junnplus: > Listen `:0` could make this port open to the public.

At the end of this function, the listener will be closed directly, ":0" or "127.0.0.1:0" are the same.

gaocegege: > BTW, why not pass a `Listener` to the ssh client directly? Which is more common in golang?
> 
> @gaocegege

Actually, the ssh client code is copied from https://github.com/okteto/okteto.

I think passing Listener also works but okteto's package passes the port.

gaocegege:
I am not sure if the problem is related to arch Linux or your localhost settings. But I think we can merge it since it does not have side effects.

WDYT @Xuanwo ?

gaocegege:
I am merging the PR, but feel free to leave comments here if you are still interested in why localhost is not set to 127.0.0.1

Xuanwo:
> I am not sure if the problem is related to arch Linux or your localhost settings.

Yep, it does. But there are some environments that don't set `localhost` correctly, so change `localost` will make their lives easier. So I will this PR an LGTM.

gaocegege:
@all-contributors Please add @junnplus for the code contribution.

Issue #347, opened 2022-06-15
feat: Homebrew support


rudeigerc:
Since `envd` supports both macOS and Linux, it would be better for us to support Homebrew as an alternative choice for installation. `goreleaser` also supports generating and publishing Homebrew Taps [by default](https://goreleaser.com/customization/homebrew/).

References:
- https://github.com/weaveworks/footloose/blob/master/.goreleaser.yml#L29-L39
- https://github.com/weaveworks/homebrew-tap

popfido:
install envd through pip does seems quite strange. Do we need a separate repo to manage the tap?

gaocegege:
@popfido We support pip since data scientists always use it. Thus we build python wheels although envd is a golang project.

Do you prefer `brew install` or `apt install`? I am thinking if we are going in the right direction about installation.

/cc @VoVAllen @kemingy

gaocegege:
> Do we need a separate repo to manage the tap?

I am not familiar with brew. I am a Ubuntu user.

I think the goreleaser can take care of tap and we do not need to maintain them in the repository. Correct me if I am wrong. cc @rudeigerc

rudeigerc:
@popfido @gaocegege 

We do need a tap repository for maintaining Homebrew Formula. goreleaser would handle this for us. It supoorts auto-generating the formula and pushing to the tap repo.

gaocegege:
One of the reasons why I do not like macOS is homebrew :smile:

kemingy:
> One of the reasons why I do not like macOS is homebrew 😄

Same here. I manually install everything on macOS to avoid homebrew breaking the whole system.

BTW, we can also support conda. (not sure about nix 👀

gaocegege:
nix is too hardcore. I do not think we need to support it since our target users may do not even know about it.

popfido:
@gaocegege If in Linux, seldom does a DS use brew to install something, but it is even bad to use pip. Coz it's often a hard job for DS to identify whether the installation is in a system python env or a virtual python or in conda's default env, etc. Placing envd in `/usr/local/bin` by package manager in various system might be more suitable.

gaocegege:
SGTM. It is reasonable.

kemingy:
It's okay if they don't know which `pip` they are using because we don't depend on Python versions. It's just a wrapper to call the binary. Of course, it's better to install with `pipx`.

rudeigerc:
[...]

PR #389, opened 2022-06-16
feat(release): Support Homebrew in goreleaser

rudeigerc:
Signed-off-by: Yuchen Cheng <rudeigerc@gmail.com>

Related to #347

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -13,7 +13,7 @@ builds:
     main: ./cmd/envd
     binary: envd
     ldflags:
-      - -s -w 
+      - -s -w
       - -X github.com/tensorchord/envd/pkg/version.version={{ .Version }}
       - -X github.com/tensorchord/envd/pkg/version.buildDate={{ .Date }}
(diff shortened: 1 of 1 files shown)


.goreleaser.yaml
```diff
@@ -89,3 +89,21 @@ docker_manifests:
...
+brews:
+  - name: envd
+    ids:
+    - envd
+    tap:
+      owner: tensorchord
+      name: homebrew-tap
+    commit_author:
+      name: Ce Gao
+      email: cegao@tensorchord.ai
```
gaocegege, line 101: Is it possible to use org as the commit author here?

gaocegege (approved):
LGTM with nits

.goreleaser.yaml
rudeigerc: Sure. This section just sets the username and email of git in commit messages.

gaocegege: Then let's use TensorChord and envd-maintainers@tensorchord.ai here

rudeigerc: Got it.

rudeigerc:
Closes #347

PR #405, opened 2022-06-16
feat: Support configuring CRAN mirror for R environment

terrytangyuan:
Signed-off-by: terrytangyuan <terrytangyuan@gmail.com>


gaocegege:
LGTM! We may need to think if we need to support more languages e.g. julia.

terrytangyuan:
> LGTM! We may need to think if we need to support more languages e.g. julia.

Yep. Opened https://github.com/tensorchord/envd/issues/408 to track. There might also be demand to support multiple languages in the same environment. Opened https://github.com/tensorchord/envd/issues/407 to track that separately.

PR #414, opened 2022-06-17
chore: Build arm64 for macOS in CI wip

### LINKED ISSUE #340: bug: arm64 binaries are not available on PyPI
Currently, we only distribute amd64 on PyPI for both Mac and Linux but not arm64. 

```
    - name: Configure linux build environment
      if: runner.os == 'Linux'
      run: |
        mkdir -p bin
        mv dist/envd_linux_amd64_v1/envd bin/envd
        chmod +x bin/envd
    - name: Configure macOS build environment
      if: runner.os == 'macOS'
      run: |
        mkdir -p bin
        mv dist/envd_darwin_amd64_v1/envd bin/envd
        chmod +x bin/envd
```
We should support arm64, especially for Apple M1 users.

gaocegege:
Fix #340 

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -72,7 +72,7 @@ jobs:
         mkdir -p bin
         mv dist/envd_darwin_amd64_v1/envd bin/envd
         chmod +x bin/envd
-    - name: Build wheels on Linux
+    - name: Build wheels
       uses: pypa/cibuildwheel@v2.6.1
       env:
         CIBW_ARCHS_MACOS: "x86_64 universal2 arm64"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -52,7 +52,7 @@ jobs:
     strategy:
       matrix:
         os: [macos-10.15, ubuntu-20.04]
-    if: startsWith(github.ref, 'refs/tags/v') # only trigger when tag starts with v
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -75,7 +75,8 @@ jobs:
     - name: Build wheels
       uses: pypa/cibuildwheel@v2.6.1
       env:
-        CIBW_ARCHS: auto64
+        CIBW_ARCHS_MACOS: "x86_64 universal2 arm64"
+        CIBW_ARCHS_LINUX: "x86_64"
```
VoVAllen, line 79: What's the arch of envd? How is it configured?

gaocegege:
I think we should postpone the PR since I have no way to test if the PR works well. We need a M1 macOS to verify the wheel can be installed and the envd bin can be run successfully.

popfido:
> I think we should postpone the PR since I have no way to test if the PR works well. We need a M1 macOS to verify the wheel can be installed and the envd bin can be run successfully.

How to test it lol. I can help if possible.

gaocegege:
> How about this https://veertu.com/?

Will have a look at it. Seems helpful to us.

TaylorHere:
https://github.blog/changelog/2021-08-16-github-actions-macos-11-big-sur-is-generally-available-on-github-hosted-runners/
This post may be helpful

PR #464, opened 2022-06-19
feat(CLI): Prune cache

### LINKED ISSUE #128: feat(CLI): Prune cache (both buildkit and midi cache)
## Description

We should provide a way to clear the cache.

gaocegege:
Fix #128 

### DIFF
--- a/pkg/app/prune.go
+++ b/pkg/app/prune.go
@@ -1,0 +1,74 @@
+// Copyright 2022 The envd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package app
+
+import (
+	"github.com/cockroachdb/errors"
+	"github.com/tensorchord/envd/pkg/buildkitd"
+	"github.com/tensorchord/envd/pkg/home"
+	"github.com/urfave/cli/v2"
+)
+
+var CommandPrune = &cli.Command{
+	Name:  "prune",
+	Usage: "clean up build cache",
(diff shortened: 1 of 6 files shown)


pkg/app/prune.go
```diff
@@ -0,0 +1,74 @@
...
+			Name:  "keep-storage",
+			Usage: "Keep data below this limit (in MB)",
+		},
+		&cli.StringSliceFlag{
+			Name:  "filter, f",
+			Usage: "Filter records",
+		},
+		&cli.BoolFlag{
+			Name:  "all",
+			Usage: "Include internal caches",
```
VoVAllen, line 42: What's internal caches?

pkg/app/prune.go
```diff
@@ -0,0 +1,74 @@
...
+			Name:  "keep-duration",
+			Usage: "Keep data newer than this limit",
+		},
+		&cli.Float64Flag{
+			Name:  "keep-storage",
+			Usage: "Keep data below this limit (in MB)",
+		},
+		&cli.StringSliceFlag{
+			Name:  "filter, f",
+			Usage: "Filter records",
```
VoVAllen, line 38: What can be filtered?

pkg/app/prune.go
```diff
@@ -0,0 +1,74 @@
...
+	Name:  "prune",
+	Usage: "clean up build cache",
+	Flags: []cli.Flag{
+		&cli.DurationFlag{
+			Name:  "keep-duration",
+			Usage: "Keep data newer than this limit",
+		},
+		&cli.Float64Flag{
+			Name:  "keep-storage",
+			Usage: "Keep data below this limit (in MB)",
```
VoVAllen, line 34: Does it work like an LRU cache?

gaocegege: No, we do not guarantee it.

pkg/app/prune.go
gaocegege: It is the CLI flag in buildkit. You can filter records by fields like description, last used time. We should document it in the doc

pkg/app/prune.go
gaocegege: buildkit's internal cache and envd host cache like oh-my-zsh

gaocegege:
PTAL again.

VoVAllen (approved):
LGTM. Please also update the docs once merged.

PR #477, opened 2022-06-21
fix: correct comments and unify receiver names and simplify some code

hezhizhen:

### DIFF
--- a/pkg/docker/docker.go
+++ b/pkg/docker/docker.go
@@ -50,7 +50,7 @@ var (
 type Client interface {
 	// Load loads the image from the reader to the docker host.
 	Load(ctx context.Context, r io.ReadCloser, quiet bool) error
-	// Start creates the container for the given tag and container name.
+	// StartEnvd creates the container for the given tag and container name.
 	StartEnvd(ctx context.Context, tag, name, buildContext string,
 		gpuEnabled bool, numGPUs int, sshPort int, g ir.Graph, timeout time.Duration,
 		mountOptionsStr []string) (string, string, error)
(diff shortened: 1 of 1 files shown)


pkg/docker/docker.go
```diff
@@ -445,8 +435,8 @@ func (g generalClient) IsCreated(ctx context.Context, cname string) (bool, error
 	return true, nil
```
hezhizhen, line 435: Why not using container.State.Status (one of "created", "running", "paused", "restarting", "removing", "exited", or "dead")? @gaocegege

pkg/docker/docker.go
```diff
@@ -90,19 +90,17 @@ func NewClient(ctx context.Context) (Client, error) {
...
+	info, err := c.Info(ctx)
 	if err != nil {
 		return false, errors.Wrap(err, "failed to get docker info")
 	}
 	logrus.WithField("info", info).Debug("docker info")
-	if nv, ok := info.Runtimes["nvidia"]; ok {
-		return nv.Path != "", nil
-	}
-	return false, nil
+	nv, _ := info.Runtimes["nvidia"]
```
hezhizhen, line 99: https://go.dev/play/p/hyPjSxPOb35
It seems that getting a value from nil map won't panic in Go 1.18, maybe we can directly ignore the second returned value?

pkg/docker/docker.go
gaocegege: I expect to follow the implementation in docker-cli.

But cannot remember if the code is copied from docker-cli

hezhizhen: Got it.

gaocegege: Sorry I misunderstand here. The function `IsCreated` returns false if the container cannot be found. Thus I think it works with inspect. WDYT?

hezhizhen: Since the `created` is one of the status for containers, I think `exist` may reduce ambiguity. WDYT?

pkg/docker/docker.go
hezhizhen: @gaocegege PTAL.
Should I add some tests here?

pkg/docker/docker.go
gaocegege: SGTM

pkg/docker/docker.go
gaocegege: LGTM. I think it is hard to add tests here.

pkg/docker/docker.go
hezhizhen: https://github.com/tensorchord/envd/issues/478 
I'll resolve the conversation here.

gaocegege (approved):
LGTM!

gaocegege:

pkg/docker/docker.go
hezhizhen: There's a tricky way: make these two lines a function and add unit tests for the function

gaocegege: Then I prefer to keep the code readers happy

Issue #329, opened 2022-06-14
feat: Support conda/pip requirement file


VoVAllen:
## Description

Support pip install using `requirement.txt` and conda install using `env.yaml`([link](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#create-env-file-manually))

gaocegege:
SGTM. I think we should recommend users write deps in build.envd, but we should not be extreme. We should support requirements.txt.

Users also expect that we can support this.

gaocegege:
- [ ] Support requirements.txt

gaocegege:
> The syntax of requirement.txt is complex ([link](https://pip.pypa.io/en/stable/reference/requirements-file-format/)). However current implementation can cover most common usages I believe.
> 
> One question:
> 
> ```
> --find-links https://download.pytorch.org/whl/torch_stable.html
> torch==1.9.0+cu102
> ...
> ```
> 
> For example in dgl [usage](https://github.com/dmlc/dgl/blob/master/docker/install/conda_env/torch_gpu_pip.txt), will `--find-links https://download.pytorch.org/whl/torch_stable.html` be considered as a package when listing the dependency?

Yep, I think so. We just ignore comment. And every line starts without # will be considered to be a dependency.

Maybe we should mount the file into the build process, and use the file directly to install dependencies.

_Originally posted by @gaocegege in https://github.com/tensorchord/envd/issues/484#issuecomment-1162705326_

kemingy:
> * [x]  Support requirements.txt
> 
>     * [ ]  Support conda env.yaml

* conda env.yaml is supported in #674, should we close this issue?

PR #484, opened 2022-06-22
feat(lang): Support python requirements.txt in python_packages

gaocegege:
Ref #329 

### DIFF
--- a/envd/api.go
+++ b/envd/api.go
@@ -1,3 +1,17 @@
+// Copyright 2022 The envd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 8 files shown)


VoVAllen:
The syntax of requirement.txt is complex ([link](https://pip.pypa.io/en/stable/reference/requirements-file-format/)). However current implementation can cover most common usages I believe. 

One question:
For example in dgl [usage](https://github.com/dmlc/dgl/blob/master/docker/install/conda_env/torch_gpu_pip.txt), will `--find-links https://download.pytorch.org/whl/torch_stable.html` be considered as a package when listing the dependency?

gaocegege:
> The syntax of requirement.txt is complex ([link](https://pip.pypa.io/en/stable/reference/requirements-file-format/)). However current implementation can cover most common usages I believe.
> 
> One question:
> 
> ```
> --find-links https://download.pytorch.org/whl/torch_stable.html
> torch==1.9.0+cu102
> ...
> ```
> 
> For example in dgl [usage](https://github.com/dmlc/dgl/blob/master/docker/install/conda_env/torch_gpu_pip.txt), will `--find-links https://download.pytorch.org/whl/torch_stable.html` be considered as a package when listing the dependency?

Yep, I think so. We just ignore comment. And every line starts without # will be considered to be a dependency.

Maybe we should mount the file into the build process, and use the file directly to install dependencies.

gaocegege:
I keep track of the problem in the original issue. Let's move forward and fix it.

PR #487, opened 2022-06-22
feat(ssh): Config ssh key permanently and globally

VoVAllen:
Change:
- When use specify a new key in `envd bootstrap`, the old key will be copied to a new one with a random name. And the entries in ssh config file is changed accordingly
- The ssh flag in up and build is kept but turned to hidden

### DIFF
--- a/pkg/app/bootstrap.go
+++ b/pkg/app/bootstrap.go
@@ -52,9 +52,8 @@ var CommandBootstrap = &cli.Command{
 		},
 		&cli.StringSliceFlag{
 			Name:    "ssh-keypair",
-			Usage:   "Manually specify ssh key pair as `publicKey,privateKey`",
+			Usage:   fmt.Sprintf("Manually specify ssh key pair as `publicKey,privateKey`. Envd will generate a keypair at %s and %s if not specified", sshconfig.GetPublicKey(), sshconfig.GetPrivateKey()),
 			Aliases: []string{"k"},
-			Value:   cli.NewStringSlice(sshconfig.GetPublicKey(), sshconfig.GetPrivateKey()),
 		},
 	},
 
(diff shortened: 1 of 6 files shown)


pkg/app/bootstrap.go
```diff
@@ -42,13 +50,81 @@ var CommandBootstrap = &cli.Command{
...
 }
 
 func bootstrap(clicontext *cli.Context) error {
+	// If not specified, generate only if key doesn't exist
+	sshKeyPair := clicontext.StringSlice("ssh-keypair")
+	if len(sshKeyPair) != 0 && len(sshKeyPair) != 2 {
+		return errors.Errorf("Invliad ssh-keypair flag")
+	} else if len(sshKeyPair) == 0 {
+
+		keyExists, err := sshconfig.DefaultKeyExists()
```
gaocegege, line 70: Is it possible, or easy, to show the default key paths in CLI flag?

pkg/app/bootstrap.go
```diff
@@ -78,3 +154,7 @@ func bootstrap(clicontext *cli.Context) error {
 	}
 	return nil
 }
+
+func FileExists() {
```
gaocegege, line 158: I am wondering if we can remove it.

pkg/util/fileutil/namegenerator.go
```diff
@@ -0,0 +1,869 @@
+// Copyright 2022 The moby Authors
```
gaocegege, line 1: This should be the second line. Keep envd the first line.

pkg/app/bootstrap.go
VoVAllen: Added

gaocegege (approved):
LGTM :+1:
/lgtm

PR #506, opened 2022-06-25
chore: remove useless ssh_config file

zdyxry:


gaocegege:
Thanks for the fix! I think we forget to remove it when refactoring the code base.

gaocegege:
@al-contributors Please add @zdyxry for code contribution.

PR #522, opened 2022-06-28
fix: uid corrupted when run envd by root user

aseaday:

### DIFF
--- a/pkg/lang/ir/system.go
+++ b/pkg/lang/ir/system.go
@@ -118,19 +118,40 @@ func (g *Graph) compileBase() llb.State {
 	} else {
 		base = g.compileCUDAPackages()
 	}
+	var res llb.ExecState
 	// TODO(gaocegege): Refactor user to a separate stage.
-	res := base.
-		Run(llb.Shlex(fmt.Sprintf("groupadd -g %d envd", g.gid)),
-			llb.WithCustomName("[internal] create user group envd")).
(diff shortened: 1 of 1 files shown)


aseaday:
It is a workaound for root users.
We need a better design for envd user and its use in image building and user use.

pkg/lang/ir/system.go
```diff
@@ -118,6 +118,9 @@ func (g *Graph) compileBase() llb.State {
 	} else {
 		base = g.compileCUDAPackages()
 	}
+	if g.uid == 0 {
+		g.uid = 1000
```
gaocegege, line 122: I used to think it works. But I am not sure since the working dir mounted in the container is also owned by root user.

Could you please test if it works?

aseaday: Let me do more test on it.  I overlooked there is a more step mount files on the directory.

gaocegege:
@aseaday Hi, is this PR ready to review?

aseaday:
> @aseaday Hi, is this PR ready to review?

I need do some tests in conda install and different shell use. Later today.

gaocegege:
No hurry. Thanks! 

Please at me @gaocegege  if it is ready to review.

aseaday:
PTAL

aseaday:
屏幕截图 2022-06-30 012157

gaocegege (approved):
LGTM :+1:
/lgtm

gaocegege:
@all-contributors Please add @aseaday for the code contribution.

PR #303, opened 2022-06-13
feat: Add envd-server runtime proposal

gaocegege:
Ref #179 

Proposal preview: https://github.com/gaocegege/envd/blob/proposal/docs/proposals/20220603-kubernetes-vendor.md

PR #533, opened 2022-06-30
feat(docker): Add entrypoint and ports in image config

gaocegege:
Ref #303 

This PR is to add entrypoints and exposed ports into the image config.

EXPOSE and ENTRYPOINT instruction doesn’t actually exposes port and sets entrypoint to the external environment, it acts as documentation between person who’s writing Dockerfile and person who’s running Dockerfile.


gaocegege:
> So the port/entrypoint configuration is not exposed to the user yet?

Yep, they are exposed to the people who run the images. But if you are the end-user, you do not need to deal with these things

And, we can remove port in `config.jupyter`. We will use a random host port to forward.

VoVAllen:
> > So the port/entrypoint configuration is not exposed to the user yet?
> 
> Yep, they are exposed to the people who run the images. But if you are the end-user, you do not need to deal with these things
> 
> And, we can remove port in `config.jupyter`. We will use a random host port to forward.

Is it still configureable？

gaocegege:
No, the port does not work after the PR. Actually, there is no need to specify the port since we can choose a random one, just like ssh does.

gaocegege:
Please see the change https://github.com/tensorchord/envd/pull/533/commits/8f384cc39814c7da0bdbf152033cfcb7518f96b9

@VoVAllen

Issue #478, opened 2022-06-21
feat: replace `IsCreated` with `Exists` for `Client` interface from package `docker`


hezhizhen:
## Description

`created` is one of the status for containers and the `IsCreated` method in `Client` interface checks if the container exists (no matter if it is running or not running properly).

To reduce ambiguity, I think it's better to add a new method `Exists` to the `Client` interface and replace `IsCreated` with it.

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

zhyon404:
I would like to help with this.

gaocegege:
/assign @zhyon404 

Thanks! Do not hesitate to contact us on Discord if you have any problems.

PR #558, opened 2022-07-06
feat: replace IsCreated with Exists for Client interface from package docker

zhyon404:
fix https://github.com/tensorchord/envd/issues/478

### DIFF
--- a/pkg/docker/docker.go
+++ b/pkg/docker/docker.go
@@ -59,7 +59,7 @@ type Client interface {
 	StartBuildkitd(ctx context.Context, tag, name, mirror string) (string, error)
 
 	IsRunning(ctx context.Context, name string) (bool, error)
-	IsCreated(ctx context.Context, name string) (bool, error)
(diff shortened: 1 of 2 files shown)


gaocegege:

gaocegege:
This PR fixes #478

gaocegege (approved):
LGTM :+1:
/lgtm

pkg/docker/docker.go
```diff
@@ -59,7 +59,7 @@ type Client interface {
 	StartBuildkitd(ctx context.Context, tag, name, mirror string) (string, error)
 
 	IsRunning(ctx context.Context, name string) (bool, error)
-	IsCreated(ctx context.Context, name string) (bool, error)
```
hezhizhen, line 62: It's better to add a new method `Exists` and mark `IsCreated` deprecated.

gaocegege: We are still in alpha, thus I think it also works although we remove it directly.

There is no project that relies on us, currently.

PR #566, opened 2022-07-07
bug: Fix github action

### LINKED ISSUE #565: bug:  Invalid workflow file: .github/workflows/release.yml#L57
## Description

Got emails from GitHub when merging commits:

* .github/workflows/release.yml: No jobs were run

Check https://github.com/tensorchord/envd/actions/runs/2628201836

---

VoVAllen:

Fix #565

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -48,13 +48,13 @@ jobs:
           if-no-files-found: error
   pypi_publish:
     needs: goreleaser
-    if: github.repository == 'tensorchord/envd'
+    # only trigger on main repo when tag starts with v
+    if: github.repository == 'tensorchord/envd' && startsWith(github.ref, 'refs/tags/v')
     runs-on: ${{ matrix.os }}
     timeout-minutes: 20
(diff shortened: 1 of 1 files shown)


kemingy:
Do we need to run `image_publish` for all the commits? It seems really slow. I guess it only needs to run for tagged commit?

gaocegege:
> Do we need to run `image_publish` for all the commits? It seems really slow. I guess it only needs to run for tagged commit?

SGTM. Could you please update the PR? @VoVAllen

VoVAllen:
Actually it will be only triggered with tag push
```
on:
  push:
    tags:
      - 'v*'
  pull_request:
    paths:
    - '.github/workflows/release.yml'
    - '.goreleaser/'
    - '.goreleaser.yaml'
```

Adding `startsWith(github.ref, 'refs/tags/v')` in pypi_publish is because pypi index can only accept formal tag version name. This is used to avoid potential error when modifying other part of `release.yml`

gaocegege:
SGTM

Issue #527, opened 2022-06-29
feat(lang): Support TensorBoard


gaocegege:
## Description

```python
def build():
  config.tensorboard(port="8888", log_dir="./logs")
```

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

terrytangyuan:
I am not sure if we should support framework-specific functions. Perhaps we should make it generic to expose any services instead.

aseaday:
I think there are some questions to discuss:
- Should users care about the port? if we listen on the host's interface, user A use port 8888 but user B used it already.
- Which interface should be listening ,the container or the host.
- Which address? Loopback or external ip?
We should anwser the above questions to design the service function.
In my opinon, we should sperate the service and the service provider(such as Service and Deployment in K8S).

```python
class JuypterSvc(Service):
  port = 8000
  interfaceLoc = container
  ip = 0.0.0.0
def build():
  config.juypter(service=JuypterSvc)
```

gaocegege:
> Should users care about the port? if we listen on the host's interface, user A use port 8888 but user B used it already.

I think in the best case, users do not need care about the port. He can use the endpoint provided by envd to visit these services.

> Which interface should be listening ,the container or the host.

Currently, we keep port in the container open and use docker port forwarding to forward container ports to host.

> Which address? Loopback or external IP?

Since we forward container port to host, thus localhost can be used here. I am not sure if it works for WSL @VoVAllen

VoVAllen:
WSL will automatically make ports available to host os (windows).

We definitely need a way for users to easily extend their use cases. However, the grammar may need more discussions.

kemingy:
Since `envd` already supports `expose`, `daemon`, `mount`, I think it can be done in the following way:

```py
def tensorboard():
    io.mount(src="/var/log/tensorboard", dest="./logs")
    runtime.daemon(commands=[
        ["tensorboard", "--logdir", "./logs", "--port", "8888"],
    ]),
    runtime.expose(envd_port=8888, host_port=8888, service="tensorboard")
```

Shall we close this issue? @gaocegege

aseaday:
All we need is code snipplet

gaocegege:
Ref #91 

I think we can close it after the doc is updated, WDYT.

gaocegege:
In the future, I think we should have the mechanism to make these starlark funcs built-in in envd.

gaocegege:
[...]

Issue #157, opened 2022-05-18
feat(lang): v2 language new design


gaocegege:
## Description

envd is designed to be the development environment for training. But I am thinking if we should support serving.

There are a few things we need to confirm:

- The pain points of environment management in serving
- The current experience
- How to integrate with serving servers
- How to integrate with model monitoring

kemingy:
The serving environment has the same problem: it's hard for the data scientists to write a Dockerfile with the **best** practice and integrate it with the CI/CD pipeline.

Usually, it's offered by the AI infra team, but they may not be able to review all the changes.

Requirements:
* non-root
* CUDA runtime & version (may need to run `ldconfig`)
* required Python packages (installed from pip or conda)
* environments (optional, like `OMP_NUM_THREADS`, `MKL_NUM_THREADS`, `CUDA_VISIBLE_DEVICES`, etc.)
* where to get the model data
* entrypoint (`tini` or some other tools)

aseaday:
Another important point about serving is `log`. I think we should consider it .

gaocegege:
Should we introduce a new CLI to support the serve?

Maybe envd deploy or something else.

kemingy:
To make it clean, here is a diagram of the current dev env build graph:

![image](https://user-images.githubusercontent.com/12974685/202422038-298c9dc3-9460-473f-bdc2-eade82b39625.png)

kemingy:
> Another important point about serving is `log`. I think we should consider it .

Which kind of logger/collectors are you using? I used to try the GCP StackDriver, users only need to log stdout/stderr in JSON format.

> Should we introduce a new CLI to support the serve?
> 
> Maybe envd deploy or something else.

SGTM.

aseaday:
Maybe we could design a interface such as:
```
runtime.log(driver="logstash", ...)
```

VoVAllen:
I think in serving scenario, logs are handled by k8s. Such as a deamonset, or a sidecar container to send the logs to remote API endpoint.
Is there anything special that envd needs take care of? I think the only thing we need to support here is the log file path?

kemingy:
I'm trying to make it compatible with the current interface.

A draft for the envd file:

```starlark
envdlib = include("https://github.com/tensorchord/envdlib")

requirements = [
    "torch --extra-index-url https://download.pytorch.org/whl/cpu",
    "torchvision",
    "numpy",
    "mosec",
    "msgpack",
    "Pillow",
]

[...]

PR #568, opened 2022-07-07
feat(lang): Add proposal for expose

aseaday:

### DIFF
--- a/docs/proposals/20220707-service-primitive.md
+++ b/docs/proposals/20220707-service-primitive.md
@@ -18,7 +18,7 @@ In the future, we should also keep the possibility for remote or cloud access.
     - Linux/Mac
     - wsl access from the windows
 
-## Proposal
+## API
 ```
 expose(local_port=<LPORT>, host_port=<RPORT>, svc=<SERVICE_NAME>)
 ```
--- a/docs/proposals/20220707-service-primitive.md
+++ b/docs/proposals/20220707-service-primitive.md
@@ -41,7 +41,7 @@ def build():
     # if you do not need to set up password.
     config.jupyter(password="password")
 
-def server():
+def serve():
     pass
 ```
 
@@ -54,7 +54,7 @@ def build():
     # if you do not need to set up password.
     config.jupyter(password="password")
 def serve():
-    host.bind(env_port=presets.juypter.port,host_port=1234)
+    host.bind(env_port=presets.juypter.port,host_port=1234) # The bind should judge the wsl automatically
 ```
 
 for the cloud user scenario:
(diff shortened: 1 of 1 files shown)


VoVAllen (commented):
I'm not sure whether we want to expose k8s related function to user now. Probably we want a unified interface such as `bind` or `expose` and will have different implementation based on different contexts proposed in https://github.com/tensorchord/envd/pull/303. @gaocegege What do you think?

docs/proposals/20220707-service-primitive.md
```diff
@@ -0,0 +1,87 @@
...
+
+
+```python
+def build():
+    base(os="ubuntu20.04", language="python")
+    # Use `config.jupyter(password="")` 
+    # if you do not need to set up password.
+    config.jupyter(password="password")
+def serve():
+    cloud.k8s.bind(env_port=presets.juypter.port,host_port=1234) # It would use port forward
```
gaocegege, line 70: It relates to #303 

I am not sure if we should introduce the `cloud` package here. But I do think we need a mechanism like `host.bind` to support other services like #527 TensorBoard.

kemingy: * The name `serve` has ambiguity with `serving` https://github.com/tensorchord/envd/issues/157

aseaday: > * The name `serve` has ambiguity with `serving` [discussion(lang): Should we support serving? #157](https://github.com/tensorchord/envd/issues/157)

I think `expose` is a more appropriate keyword.

aseaday:
@gaocegege @VoVAllen @kemingy Thanks for your suggestions, I am going to:
- Rename `serve` to `expose`, because I think it is a exposing helper in the development process and `serve` may cause problems.
- Remove cloud package for now and only keep `host.bind`

aseaday:
PTAL

docs/proposals/20220707-service-primitive.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Goals
+- http/tcp 
+- OS support:
+    - Linux/Mac
+    - wsl access from the windows
+
+## API
+```
+expose(local_port=<LPORT>, host_port=<RPORT>, svc=<SERVICE_NAME>)
```
gaocegege, line 23: How do we know what command should be run to launch the service in envd?

aseaday: I am not sure wether we should take of this in `expose`. I still think we could't see the envd as a docker which means:
1. A container which has a entrypoint to run as a service
2. If we provide some functions to run something in the near future, It'd better be like `run(juypter)` according the presets we defined for users.
So I think `expose` can be seprated from the application.

gaocegege: Then we need a mechanism to run some user-defined commands in the container. `run()` does not work in this scenario since it is just the docker `RUN` command, and is launched during the build process, instead of the runtime. We need to run the exposed services in runtime.

aseaday: > Then we need a mechanism to run some user-defined commands in the container. `run()` does not work in this scenario since it is just the docker `RUN` command, and is launched during the build process, instead of the runtime. We need to run the exposed services in runtime.

Yes. We need a keyword to run some applications like juypter during runtime. So there is going to be another proposal which provide the users to start some applications and maybe `start` like:
```python
def build():
    install.python_packages(name = [
        "numpy",
        "juypter"
    ])
    shell("zsh")
    start(app.juypter)
```

gaocegege: LGTM.

gaocegege (approved):

kemingy (approved):
LGTM. Maybe we can adjust the PR title?

aseaday:
> LGTM. Maybe we can adjust the PR title?

Do you have any ideas or what format is better

aseaday:
> LGTM. Maybe we can adjust the PR title?

Sry, I got you ideas..

aseaday:
It should be ok now. @gaocegege

VoVAllen (approved):
LGTM. One thing I'm thinking about is future implementation on k8s-like context. Not sure whether host port can be a general concept. However I don't have better idea now. Thanks  for the proposal!

PR #570, opened 2022-07-07
feat: Support for build image update when exec build or up again

aseaday:

### DIFF
--- a/pkg/builder/builder.go
+++ b/pkg/builder/builder.go
@@ -120,7 +120,50 @@ func (b generalBuilder) NumGPUs() int {
 	return ir.NumGPUs()
 }
 
+// Always return updated when met error
+func (b generalBuilder) CheckDepsFileUpdate(ctx context.Context, tag string, deps []string) (bool, error) {
+	dockerClient, err := docker.NewClient(ctx)
+	if err != nil {
+		return true, err
+	}
+	image, err := dockerClient.GetImage(ctx, tag)
+	if err != nil {
+		return true, err
+	}
+	imageCreatedTime := image.Created
+
+	latestTimestamp := int64(0)
+	for _, dep := range deps {
+		file, err := os.Stat(dep)
+		if err != nil {
+			return true, err
+		}
+		modifiedtime := file.ModTime().Unix()
+		// Only needt o use the latest modified time
+		if modifiedtime > latestTimestamp {
+			latestTimestamp = modifiedtime
+		}
+	}
+	if latestTimestamp > imageCreatedTime {
+		return true, nil
+	}
+	return false, nil
+}
+
 func (b generalBuilder) Build(ctx context.Context, pub string) error {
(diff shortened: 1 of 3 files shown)


aseaday:
It is not a formal PR. I found some problems when check updates between image and manifest files.
if the manifest changed like A->B->A, the newest image woule always at a old time as the first A maybe caused by the cache system.  So when you execute `envd up`, it would always build image again according the current implementation: It just compare the timestamp between image and manifest.

We could only accept this problem or we introduce a hash of the manifest and label it on the image:

Hash(manifest.AST + pubkey + config_file)

pkg/builder/builder.go
```diff
@@ -120,7 +120,50 @@ func (b generalBuilder) NumGPUs() int {
 	return ir.NumGPUs()
 }
 
+// Always return updated when met error
+func (b generalBuilder) CheckDepsFileUpdate(ctx context.Context, tag string, deps []string) (bool, error) {
```
gaocegege, line 124: Fancy implementation!

I am not sure how docker deals with the ImageCreateTime in our case. Did you test the logic locally?

pkg/builder/builder.go
```diff
@@ -120,7 +120,50 @@ func (b generalBuilder) NumGPUs() int {
...
+	}
+	if latestTimestamp > imageCreatedTime {
+		return true, nil
+	}
+	return false, nil
+}
+
 func (b generalBuilder) Build(ctx context.Context, pub string) error {
+	depsFiles := []string{
+		pub,
```
gaocegege, line 155: I am not sure what `pub` is, could you please explain more about it?

pkg/builder/builder.go
aseaday: I test it. It seems that depends the tarball from the buildkit. So it could be cached if llb didn't change.

pkg/builder/builder.go
aseaday: The pub key is filepath "/home/aseaday/.config/envd/id_rsa_envd.pub". I think It can be changed if users bootstrap again. So it need to be added into depsFiles.

gaocegege (approved):
LGTM!

cc @VoVAllen 

Please have a look since it is related to ssh part.

VoVAllen (approved):
LGTM! Nice check on the dependencies!

gaocegege:
> It is not a formal PR but for your own exp.
> 
> I found some problems when check updates between image and manifest files. if the manifest changed like A->B->A, the newest image woule always at a old time as the first A maybe caused by the cache system. So when you execute `envd up`, it would always build image again according the current implementation: It just compare the timestamp between image and manifest.
> 
> We could only accept this problem or we introduce a hash of the manifest and label it on the image:
> 
> Hash(manifest.AST + pubkey + config_file)

Maybe we should document it in envd docs.

Issue #572, opened 2022-07-08
bug: ERROR exporting to oci image format "invalid reference format: repository name must be lowercase"


kemingy:
## Description

Got an error when the dir name is uppercase.

## Reproduction

```sh
mkdir -p Workspace
cd Workspace
# add envd file
envd up
```

## Additional Info

error msg

```
❯ envd up                                                                                                    (base)
WARN[2022-07-08T10:49:06+08:00] start the created contrainer envd_buildkitd   container=envd_buildkitd driver=docker-container image="docker.io/moby/buildkit:v0.10.3" socket=envd_buildkitd
v0.10.3: Pulling from moby/buildkit
Digest: sha256:0dc312b04eac1b44cd2cad566deb1e886c753109208affbbec8384f381ff7f38
Status: Downloaded newer image for moby/buildkit:v0.10.3
[+] ⌚ parse build.envd and download/cache dependencies 0.0s ✅ (finished)
[+] build envd environment 91.5s (13/14)
 => docker-image://docker.io/tensorchord/python:3.8-ubuntu20.04                                               56.0s
[+] build envd environment 91.5s (14/14) FINISHED
 => docker-image://docker.io/tensorchord/python:3.8-ubuntu20.04                                               56.0s
 => => resolve docker.io/tensorchord/python:3.8-ubuntu20.04                                                   15.9s
 => => sha256:8513ee3d04236b3fbce5e7779dfa9c1a509e2108d6bf37df2b7cca27ea06695e 5.49MB / 5.49MB                 2.5s
 => => sha256:dd0b398750f4b004a65c1bb9bf2c93b7a753756afaea45ac0dd29ab305661141 58.96MB / 58.96MB              18.7s
 => => sha256:3e66387987ad1b955b18ed108c3e53074f997ec0b7d93a02b11308276e0ddcf6 91.22MB / 91.22MB              36.2s
 => => extracting sha256:3e66387987ad1b955b18ed108c3e53074f997ec0b7d93a02b11308276e0ddcf6                      2.2s
 => => extracting sha256:dd0b398750f4b004a65c1bb9bf2c93b7a753756afaea45ac0dd29ab305661141                      1.5s
 => => extracting sha256:8513ee3d04236b3fbce5e7779dfa9c1a509e2108d6bf37df2b7cca27ea06695e                      0.1s
 => [internal] create user group envd                                                                          0.9s
 => [internal] create user envd                                                                                0.1s
[...]

PR #573, opened 2022-07-08
Fix: image tag normalized to docker spec

aseaday:
- rename the image tag do lowcased if reference has some uppers
- throw error if the directory  just a digest [a-z0-9]{64}


aseaday:
https://github.com/tensorchord/envd/issues/572

gaocegege (approved):

Especially for the test!

PR #579, opened 2022-07-08
chore: update envd destroy cmd log msg

kemingy:

Make sure users get useful messages.


VoVAllen (approved):
Nice catch!

PR #590, opened 2022-07-11
chore: Improve error wrapping and checking

terrytangyuan:
This fixes the following error lint checks:

```
comparing with == will fail on wrapped errors. Use errors.Is to check for a specific error
type assertion on error will fail on wrapped errors. Use errors.As to check for specific errors
```

Signed-off-by: Yuan Tang <terrytangyuan@gmail.com>


gaocegege (approved):
LGTM!

PR #591, opened 2022-07-11
ci: Replace lint checks with pre-commit and golangci-lint linters

terrytangyuan:
We currently use `git diff --cached --exit-code`, which can be replaced by `trilom/file-changes-action`. The lint checks such as `vet`, `fmt`, `mod tidy,` are available in `golangci-lint`. It will be easy to add other linters in golangci-lint in the future. 

I can add instructions to the docs later but it's as simple as `pre-commit run` and will automatically run when you `git commit` your changes once the hooks have been installed via `pre-commit install`.


gaocegege (approved):

Issue #581, opened 2022-07-08
feat: Support python version


VoVAllen:
## Description

Currently the python version is fixed at https://github.com/tensorchord/envd/blob/main/pkg/lang/ir/system.go#L114

Related: https://github.com/tensorchord/envd/pull/567
---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

kemingy:
BTW, installing `Python` through `apt` is not a good practice.

* https://pythonspeed.com/articles/official-python-docker-image/
* https://pythonspeed.com/articles/importerror-docker/

gaocegege:
> BTW, installing `Python` through `apt` is not a good practice.
> 
> * https://pythonspeed.com/articles/official-python-docker-image/
> * https://pythonspeed.com/articles/importerror-docker/

Could you please explain more about it? I did not find why not to use apt to install python in these two articles.

kemingy:
> Could you please explain more about it? I did not find why not to use apt to install python in these two articles.

* apt source is not the latest one
  * https://github.com/tensorchord/envd/issues/580#issuecomment-1178835413  is still 3.8.10 while the latest one is 3.8.13, check https://www.python.org/downloads/
* https://frostming.com/2022/03-27/python-on-debian/  apt source is special

gaocegege:
Maybe we can use conda to manage python in the envd environment. We create a conda env in every envd image fs, and use this env to launch the shell program. Users always use conda python instead of the system python.

aseaday:
I think in most cases, conda env manager is a enough solution.

aseaday:
https://github.com/tensorchord/envd/pull/602 solved this problem in most cases. Maybe more documentation needed.

popfido:
conda is a good solution, but conda's environment solving speed is horrable. Maybe we can also support `mamba` or `micromamba`

aseaday:
It is a good idea to use mamba because I used it too. 
IMO, we'd better add a extra option for user to use. `conda` is popular at least for our users.

Issue #580, opened 2022-07-08
bug: default `python` in the container is incorrect if `zsh` is not used


kemingy:
## Description

`build.envd`
```py
def build():
    base(os="ubuntu20.04", language="python3")
    install.python_packages(name=["ipython"])
```

## Reproduction

```sh
envd up
```

```console
envd@814312785d13:~/examples$ which python
/opt/conda/bin/python
envd@814312785d13:~/examples$ which python3
/usr/bin/python3
envd@814312785d13:~/examples$ which pip
/usr/bin/pip
envd@814312785d13:~/examples$ ipython
bash: ipython: command not found
envd@814312785d13:~/examples$ pip show ipython
WARNING: The directory '/home/envd/.cache/pip' or its parent directory is not owned or is not writable by the current user. The cache has been disabled. Check the permissions and owner of that directory. If executing pip with sudo, you may want sudo's -H flag.
Name: ipython
Version: 8.4.0
Summary: IPython: Productive Interactive Computing
Home-page: https://ipython.org
Author: The IPython Development Team
Author-email: ipython-dev@python.org
License: BSD
Location: /home/envd/.local/lib/python3.8/site-packages
Requires: pexpect, pickleshare, decorator, setuptools, stack-data, pygments, prompt-toolkit, matplotlib-inline, backcall, jedi, traitlets
Required-by:
envd@814312785d13:~/examples$ echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/opt/conda/bin:/usr/local/julia/bin
```

## Additional Info

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritise the issues with the most 👍.

kemingy:
I highly recommend adding some tests for this one. Since this should be the fundamental feature we need to guarantee.

kemingy:
Comparing to `zsh` config:

`build.envd`

```py
def build():
    base(os="ubuntu20.04", language="python3")
    install.python_packages(name=["ipython"])
    shell("zsh")
```

```console
(envd) ➜  examples which python
python: aliased to python3
(envd) ➜  examples which python3
/usr/bin/python3
(envd) ➜  examples which pip
/usr/bin/pip
(envd) ➜  examples which pip3
/usr/bin/pip3
(envd) ➜  examples ipython
Python 3.8.10 (default, Mar 15 2022, 12:22:08)
Type 'copyright', 'credits' or 'license' for more information
[...]

PR #602, opened 2022-07-13
fix: set conda as the only python provider

lalawuu:

https://github.com/tensorchord/envd/issues/581
https://github.com/tensorchord/envd/issues/580

### DIFF
--- a/base-images/python3.8-ubuntu20.04.Dockerfile
+++ b/base-images/python3.8-ubuntu20.04.Dockerfile
@@ -25,7 +25,7 @@ RUN apt-get update && \
     bzip2 ca-certificates libglib2.0-0 libsm6 libxext6 libxrender1 mercurial \
     procps subversion wget \
     # envd dependencies
-    curl openssh-client git tini sudo python3-pip zsh vim \
+    curl openssh-client git tini sudo zsh vim \
     && rm -rf /var/lib/apt/lists/*
 
 # Leave these args here to better use the Docker build cache
--- a/base-images/python3.8-ubuntu20.04-cuda11.6.Dockerfile
+++ b/base-images/python3.8-ubuntu20.04-cuda11.6.Dockerfile
@@ -76,7 +76,7 @@ RUN apt-get update && \
     bzip2 ca-certificates libglib2.0-0 libsm6 libxext6 libxrender1 mercurial \
     procps subversion wget \
     # envd dependencies
-    curl openssh-client git tini sudo python3-pip zsh vim \
+    curl openssh-client git tini sudo zsh vim \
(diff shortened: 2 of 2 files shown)


gaocegege (commented):
Should we remove system python in the base image?

aseaday:
> Should we remove system python in the base image?

I think it is ok.

base-images/python3.8-ubuntu20.04.Dockerfile
```diff
@@ -25,7 +25,7 @@ RUN apt-get update && \
     bzip2 ca-certificates libglib2.0-0 libsm6 libxext6 libxrender1 mercurial \
     procps subversion wget \
     # envd dependencies
-    python3 curl openssh-client git tini sudo python3-pip zsh vim \
+    curl openssh-client git tini sudo python3-pip zsh vim \
```
gaocegege, line 28: Also remove the python3-pip here please

aseaday: I split it in two commit. sry

pkg/lang/ir/python.go
```diff
@@ -24,6 +24,17 @@ import (
...
 
+func (g Graph) GetAppropriatePythonVersion() (string, error) {
+	version := *g.Language.Version
+	if version == "3" || version == "" {
+		return defaultVersion, nil
+	}
+	if strings.HasPrefix(version, "3.") {
+		return version, nil
+	}
+	return "", errors.Errorf("python version %s is not supported", version)
```
gaocegege, line 35: We do not support Python 2.x, does it work for you? @VoVAllen @kemingy

base-images/python3.8-ubuntu20.04.Dockerfile
gaocegege: LGTM!

aseaday: Should I also rename the Dockerfile? maybe python-ubuntu20.04? But I think I need help to fix other hardcode such as image base in the code and CI process.

pkg/lang/ir/python.go
aseaday: I forget to discuss with others. But I do think it is not a good idea to support python2. >_<!

gaocegege:
Please fix the lint issues

pkg/lang/ir/python.go
kemingy: LGTM

kemingy (approved):

pkg/lang/ir/python.go
VoVAllen: LGTM. Python 2.x should be deprecated in all ways

PR #604, opened 2022-07-14
feat(WSL): Add ssh config entry to Windows ssh config if using WSL

VoVAllen:
Add entry to windows ssh config also to enable user directly connect to envd environment from windows. WSL's port is automatically mapped to windows, so the host and port is shared. Key need to be manually copied to windows


gaocegege:
Should we update doc about WSL support?

VoVAllen:
Tracked in https://github.com/tensorchord/envd-docs/issues/76

PR #640, opened 2022-07-20
chore: add e2e test helper for container use spec

aseaday:
This PR is for a high level tests focuse on:
- Users could enter into the container and execute the script or comands correctly 
- Robust tests  such as old version compatibility
- Provide some DSL for write e2e test easily


aseaday:
Write a e2e test code like:
```
package e2e

import (
	. "github.com/onsi/ginkgo/v2"
	. "github.com/onsi/gomega"
)

var _ = Describe("e2e quickstart", Ordered, func() {
	exampleName := "quick-start"
	BeforeAll(BuildImage(exampleName))
	BeforeEach(RunContainer(exampleName))
	It("execute python demo.py", func() {
		Expect(example(exampleName).Exec("python demo.py")).To(Equal("[2 3 4]"))
	})
	AfterEach(DestoryContainer(exampleName))
	AfterAll(RemoveImage(exampleName))
})
```

e2e/quick_start_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+package e2e
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("e2e quickstart", Ordered, func() {
+	exampleName := "quick-start"
```
gaocegege, line 23: Maybe we should place the example in testdata/.

gaocegege (commented):
Could you please move the test cases in pkg/app into the e2e, then update the CI.yaml?

gaocegege:
Thanks for the contributions! I like the DSL!

e2e/quick_start_test.go
aseaday: LGTM

aseaday: I think I could move the extra `envd-quick-start` to testdata. And other tests shoule be wrote to cover the examples

gaocegege: examples may be complex. Personally, I prefer to test different features in different test cases. WDYT

aseaday: LGTM

aseaday: what do you mean ` then update the CI.yaml?`

Makefile
```diff
@@ -155,6 +155,7 @@ test: generate  ## Run the tests
 
 e2e-test: generate
 	@go test -race -coverpkg=./pkg/app -coverprofile=e2e-coverage.out ./pkg/app
```
gaocegege, line 157: Can we move the test suite to e2e/ directory directly?

aseaday: I move these files to the app/, But I think it is ok for now but If we have more and more spec test may not be related those commands directly. we still need a independent location to put those specs.

gaocegege: I mean move the tests in pkg/app/ to e2e/

Then all the e2e test cases are placed in the same dir.

pkg/app/e2e/build_test.go
```diff
@@ -12,14 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package app
+package e2e
```
gaocegege, line 15: I was expecting to move pkg/app to e2e/, actually. Or maybe test/e2e.

aseaday: e2e seems better after I look throught some projects.

aseaday:
I don't know why e2e test falled and I test it local

gaocegege:
Sometimes it's flakey. I triggered a new build.

gaocegege (approved):

PR #651, opened 2022-07-22
fix(build): Fix image config

### LINKED ISSUE #645: bug(build): Image config is not configured correctly
## Description

The func `imageConfig` relies on ir package to get the entrypoints/labels. But in our new design of gateway build func, the ir.DefaultImage is not initialized because we run the starlark interpretation in BuildFunc. Thus we cannot get the right config.

We should move the logic into BuildFunc with gateway client's ResolveImageConfig

```go
func (b generalBuilder) build(ctx context.Context, pw progresswriter.Writer) error {
	imageConfig, err := b.imageConfig(ctx)
	if err != nil {
		return errors.Wrap(err, "failed to get labels")
	}
	// k := platforms.Format(platforms.DefaultSpec())
	ctx, cancel := context.WithCancel(ctx)
	defer cancel()
	eg, ctx := errgroup.WithContext(ctx)
...

func (b generalBuilder) imageConfig(ctx context.Context) (string, error) {
	labels, err := ir.Labels()
	if err != nil {
		return "", errors.Wrap(err, "failed to get labels")
	}
	ports, err := ir.ExposedPorts()
	if err != nil {
		return "", errors.Wrap(err, "failed to get expose ports")
	}
	labels[types.ImageLabelContext] = b.BuildContextDir

	ep, err := ir.Entrypoint(b.BuildContextDir)
	if err != nil {
		return "", errors.Wrap(err, "failed to get entrypoint")
	}
	data, err := ImageConfigStr(labels, ports, ep)
	if err != nil {
		return "", errors.Wrap(err, "failed to get image config")
	}
	return data, nil
}
```

## Additional Info

---

gaocegege:
Fix #645 

### DIFF
--- a/examples/python-basic/build.envd
+++ b/examples/python-basic/build.envd
@@ -1,0 +1,6 @@
+def build():
+    base(os="ubuntu20.04", language="python3")
+    #config.pip_index(url = "https://pypi.tuna.tsinghua.edu.cn/simple")
+    install.python_packages([
+        "via",
+    ])
--- a/pkg/builder/build.go
+++ b/pkg/builder/build.go
@@ -18,6 +18,7 @@ import (
 	"context"
 
 	"github.com/cockroachdb/errors"
+	"github.com/moby/buildkit/exporter/containerimage/exptypes"
 	"github.com/moby/buildkit/frontend/gateway/client"
 )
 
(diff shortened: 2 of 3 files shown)


aseaday (approved):
LGTM

Issue #161, opened 2022-05-18
feat(lang): Support extension alias


gaocegege:
## Description

The extension name is hard to remember: `ms-python.python-2021.12.1559732655`. Maybe we can design a mechanism to support extensions without version tag like this `ms-python.python`

PR #652, opened 2022-07-22
fix: context create with 'use'

### LINKED ISSUE #618: bug(context): use does not work in context create command
## Description

https://github.com/tensorchord/envd/blob/main/pkg/home/context.go#L97

use here is not used.

## Additional Info

---

kemingy:

* close #618


kemingy:
* https://github.com/tensorchord/envd/issues/161#issuecomment-1129475975

It seems the API part is down.

gaocegege:
I am merging it since the e2e does not break by this PR.

Issue #673, opened 2022-07-27
feat: Declare mounting directory


VoVAllen:
## Description

Most researchers use public datasets from builtin API provided by the framework (torchvision/huggingface dataset/dgl). And those data need to be persistent, and might need to be shared between projects.

For example, using dgl, the dataset is stored in `~/dgl`. Here's the example for the Cora citation graph dataset.
The `~/.dgl` is expected to be managed by envd.
```
> tree ~/.dgl
/home/envd/.dgl
|-- config.json
|-- cora_v2
|   |-- cora_v2_dgl_graph.bin
|   |-- cora_v2_dgl_graph.pkl
|   |-- ind.cora_v2.allx
|   |-- ind.cora_v2.ally
|   |-- ind.cora_v2.graph
|   |-- ind.cora_v2.test.index
|   |-- ind.cora_v2.tx
|   |-- ind.cora_v2.ty
|   |-- ind.cora_v2.x
|   `-- ind.cora_v2.y
`-- cora_v2.zip
```

Proposal:
```
# To declare a local folder
io.mount(src="~/.envd/data/dgl", target="~/.dgl")
```

Other possible scenarios
```
# To declare envd-managed Dataset
data.persistent_volume(name=Dataset("imagenet"), target='~/.torchvision')
# To use S3 dataset
data.persistent_volume(name=S3Dataset("imagenet"), target='~/.torchvision')
```

Then envd will make this folder as mounted when starting container. 

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

gaocegege:
The name `target` confuses me. Maybe `source` is better.

VoVAllen:
How about `mount_path`? `target` here is the path inside container

gaocegege:
Then how to specify the path in the host with your design?

VoVAllen:
@gaocegege It's managed by envd. For example, save the data at `~/.envd/data`, and envd will create a unique folder if the dataset name is new, or directly mount it to the container if already exists. 

User can also specify the data path on host if needed (probably part of envd context)

VoVAllen:
MVP:

Support `io.mount(src="...", target="...")`

Optimization:
`io.mount(src=EnvdManager("dgl"), target=C.DefaultDglPath)`

PR #708, opened 2022-08-02
feat: support io.mount

VoVAllen:

Related: https://github.com/tensorchord/envd/issues/673

Support `io.mount(src=...., dest=...)` to mount host directory into container when `envd up`.

For example using dgl
```
io.mount(src="~/.envd/data/dgl", dest="~/.dgl")
```

In future we can simplify the experience, for example `io.mount(src=Envd("dgl"), dest=io.const.dgl_data_path)`


gaocegege (approved):
LGTM :+1:
/lgtm

PR #714, opened 2022-08-03
example: add a dgl GAT example

VoVAllen:

Add an example running GAT training, use `io.mount` to mount datatset


examples/dgl/train.py
```diff
@@ -0,0 +1,93 @@
+import torch
```
kemingy, line 1: Should we add the license here?

gaocegege (approved):

PR #718, opened 2022-08-03
feature: add envd top commands

aseaday:


pkg/docker/stats_type.go
```diff
@@ -0,0 +1,115 @@
+// The following code is from https://github.com/fsouza/go-dockerclient/
+// Copyright (c) 2013-2021, go-dockerclient authors
+package docker
+
+import "time"
+
+// Stats represents container statistics, returned by /containers/<id>/stats.
+// See https://goo.gl/Dk3Xio for more details.
+type Stats struct {
```
gaocegege, line 9: Do we need to maintain the struct in envd?

aseaday:
for future gpu support which go docker doesn't support now.

获取Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

pkg/docker/stats_type.go
aseaday: It is also for future support for GPU. Another point I cared about  is that the godocker repo's author has no GPG keys verified sign which could cause potienal danger.

aseaday:
image

gaocegege:

Issue #740, opened 2022-08-08
bug: failed get jupyter endpoint after destroy envd and retry to up


Xiaoaier-Z-L:
## Description
When I do as the get started, I successfully get the jupyter endpoint, but when i destroy it, and retry to up, i can not get endpoints

## Reproduce 

- modify build.envd
- envd up --detach
- envd env list
- envd destroy
- envd up --detach
- envd env list

## 
<img width="1007" alt="image" src="https://user-images.githubusercontent.com/96805673/183408539-ab137c72-adc1-4c0c-8276-d629938903e6.png">

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritise the issues with the most 👍.

gaocegege:
Thanks for the issue!

PR #741, opened 2022-08-08
The workaround to fix the label loss

aseaday:
#740


pkg/builder/builder.go
```diff
@@ -140,6 +140,9 @@ func (b generalBuilder) NumGPUs() int {
 
 func (b generalBuilder) Build(ctx context.Context, force bool) error {
 	if !force && !b.checkIfNeedBuild(ctx) {
+		if err := b.Interpret(); err != nil {
```
gaocegege, line 143: Could we add some comments to tell why we need this?

gaocegege:
LGTM :+1:
/lgtm

PR #743, opened 2022-08-08
chore: :art: refactor to use filepath correctly

zwpaper:
1. drop unnecessary filepath.Base
2. use filepath to replace path for OS safety


gaocegege (approved):
LGTM :+1:
/lgtm

Issue #446, opened 2022-06-18
bug(CLI): fail to use both path and file flag in envd up


gaocegege:
```
envd up --path ./examples/mnist -f ./examples/mnist/build_gpu.envd 
error: no such file or directory
```

We use the path dir as the cwd, thus we cannot find the file `./examples/mnist/build_gpu.envd `

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritise the issues with the most 👍.

gaocegege:
> We can recording pwd as a global 'ENV_DIR' to support both path dir and the current dir.
> 
> Is it suitable?

Could you please explain more about it?

popfido:
> > We can recording pwd as a global 'ENV_DIR' to support both path dir and the current dir.
> > Is it suitable?
> 
> Could you please explain more about it?

Just record the current working dir by using `pwd()` path as a secondary base search path when it is failed to use path dir to find the file in `-f`.

gaocegege:
Gotcha, SGTM. I think it works.

kemingy:
Is there a bug in https://github.com/tensorchord/envd/blob/53d8e13fc3cf39cd8a1c38e80d411096dd99e234/pkg/builder/util.go#L321-L338

So no matter what you pass to `-f`, it will always use the default`build.envd`. (Unless you specify the target function)

BTW, the `-f` description is "Function to execute, format `file:func`". Should we use `filepath.Base()` to extract the last element?

gaocegege:
I think it is a bug.

PR #747, opened 2022-08-09
chore: refactor the envd and up cmd

aseaday:


kemingy:
* ping me when it's ready to review
* I'm going to fix #446, this PR will make it easy (so I don't need to change both `up` and `build`

aseaday:
> * ping me when it's ready to review
> * I'm going to fix [bug(CLI): fail to use both path and file flag in envd up #446](https://github.com/tensorchord/envd/issues/446), this PR will make it easy (so I don't need to change both `up` and `build`

No problem, I am working on it and  wil have it fixed today. A headache happend to me yesterday.

kemingy:
> No problem, I am working on it and wil have it fixed today. A headache happend to me yesterday.

Take care and take your time, not an urgent task.

pkg/builder/builder.go
```diff
@@ -181,9 +176,6 @@ func (b generalBuilder) Interpret() error {
 }
 
 func (b generalBuilder) compile(ctx context.Context) (*llb.Definition, error) {
-	if err := b.Interpret(); err != nil {
```
gaocegege, line 184: Where do we run the interpreter if we removed this?

aseaday: It was moved to https://github.com/aseaday/envd/blob/75080a42d483900473f5e217f78743c26270c2d3/pkg/app/build.go#L127.
I do this because the def interpret need to be used to provide information about how to run the container. But in some cases like cache, the builder just need to interpret the manifest without needs to build the image.

gaocegege: Gotcha, while `envd run` also requires an interpretation, thus maybe we should add the logic to the run command too.

gaocegege: https://github.com/tensorchord/envd/blob/main/pkg/app/run.go#L103

Should we also need to refactor this section?

aseaday: Yes, but as for this, I am thinking about DefaultGraph or manifest's def should be accessible as a IR or  a wrap of interpreter itself. And the IR pass what builder need to builder. I am writing some docs to talk about it. I can refactor this after we all agree the design.
BTW, in this PR, I don't want a huge commit because kemingy also need a fix base on this PR.

gaocegege: SGTM

gaocegege (approved):

kemingy (approved):
LGTM thanks

Issue #670, opened 2022-07-26
feat: Support dataset integration with common frameworks


VoVAllen:
## Description

Support dataset integrations with pytorch/torchvision/huggingface-datasets. Let user can initialize data easily

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

PR #751, opened 2022-08-10
feat(data): Add support for managed dataset and provide shortcut for common framework

VoVAllen:

Syntax:
```python
io.mount(src=data.envd("dgl"), dest=data.path.dgl)
# data.envd means a data source managed by envd. Currently envd will just create a folder at ~/.envd/data/dgl if not exists and use it as source
# data.path.dgl is just an alias for ~/.dgl, which indicates the default data path of dgl. Also added the default path for huggingface dataset
```


examples/dgl/build.envd
```diff
@@ -3,12 +3,13 @@ def build():
     base(os="ubuntu20.04", language="python3")
 
     # Add the packages you are using here
-    install.python_packages(["numpy", "dgl", "torch"])
+    # install.python_packages(["numpy", "dgl", "torch"])
```
gaocegege, line 6: Should we remove the comment?

pkg/data/envdManaged.go
```diff
@@ -0,0 +1,52 @@
+// Copyright 2022 The envd Authors
```
gaocegege, line 1: The file name does not follow the golang convention, please use envd.go or datasource.go

gaocegege: https://medium.com/@kdnotes/golang-naming-rules-and-conventions-8efeecd23b68

pkg/data/util.go
```diff
@@ -0,0 +1,38 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package data
+
+// hashString computes the hash of s.
+func hashString(s string) uint32 {
```
gaocegege, line 18: Do we need to write the hash by ourselves? There are some hash packages.

gaocegege: https://github.com/google/uuid

pkg/home/data.go
```diff
@@ -0,0 +1,45 @@
...
+type dataManager interface {
+	InitDataDir(name string) (string, error)
+}
+
+func (m *generalManager) InitDataDir(name string) (string, error) {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return "", errors.Wrap(err, "failed to get user home dir")
+	}
+	newDataDir := filepath.Join(home, ".envd", "data", name)
```
gaocegege, line 34: I think we should avoid ~/.envd, maybe we keep it in ~/.cache/envd?

pkg/lang/frontend/starlark/data/util.go
```diff
@@ -0,0 +1,53 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package data
+
+import (
+	"go.starlark.net/starlark"
+
+	envdData "github.com/tensorchord/envd/pkg/data"
```
gaocegege, line 20: Camel case does not follow golang package convention, please use envddata or data

pkg/lang/frontend/starlark/io/io.go
```diff
@@ -15,6 +15,7 @@
 package io
 
 import (
+	"errors"
```
gaocegege, line 18: Please use github.com/cockroach/errors

pkg/lang/frontend/starlark/io/io.go
```diff
@@ -23,7 +24,9 @@ import (
 	"go.starlark.net/starlark"
 	"go.starlark.net/starlarkstruct"
 
+	"github.com/tensorchord/envd/pkg/lang/frontend/starlark/data"
 	"github.com/tensorchord/envd/pkg/lang/ir"
+	// envdData "github.com/tensorchord/envd/pkg/data"
```
gaocegege, line 29: Please remove the line if it is not needed.

pkg/lang/frontend/starlark/io/io.go
```diff
@@ -40,14 +43,33 @@ var Module = &starlarkstruct.Module{
...
+		if err != nil {
+			return starlark.None, err
+		}
+	} else if vs, ok := source.(starlark.String); ok {
+		sourceStr = vs.GoString()
+	} else {
+		return starlark.None, errors.New("invalid source")
+	}
+
+	// sourceStr := source.GoString()
```
gaocegege, line 72: Please remove the line if it is not needed.

gaocegege:
Close #670

pkg/home/data.go
VoVAllen: The semantic here I think is different from cache. Once a dataset is created, it is not expected to be deleted when clearing the cache. Also cache means it can be rebuilt from some other way, but I didn't see how we can rebuild user's data folder, as this is just an arbitrary folder user can play with

pkg/data/util.go
VoVAllen: Changed to starlark's string hash

pkg/home/data.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+type dataManager interface {
+	InitDataDir(name string) (string, error)
+}
+
+func (m *generalManager) InitDataDir(name string) (string, error) {
+	newDataDir := filepath.Join(m.CacheDir(), "data", name)
+	if _, err := os.Stat(newDataDir); !os.IsNotExist(err) {
```
kemingy, line 31: Can use https://github.com/tensorchord/envd/blob/8ab89b967caa104ad839a83d7d9ece186eb80918/pkg/util/fileutil/file.go#L84-L94

pkg/home/data.go
```diff
@@ -0,0 +1,41 @@
...
+		logrus.Infof("Data dir %s already exists, skipping creation", newDataDir)
+		return newDataDir, nil
+	}
+
+	err := os.Mkdir(newDataDir, 0777) // Avoid UID/GID issues
```
kemingy, line 36: `.cache/envd` is `0700`, will this be a security problem?

BTW, need to make sure this line runs after initializing the `.cache/envd`, otherwise it's better to use `os.MkdirAll`

pkg/data/envd.go
```diff
@@ -0,0 +1,54 @@
...
+	e.hostDataDir = hostDataDir
+	return nil
+}
+
+func (e *EnvdManagedDataSource) GetHostDir() (string, error) {
+	return e.hostDataDir, nil
+}
+
+func (e *EnvdManagedDataSource) Type() string {
+	return "envd managed data source"
```
kemingy, line 45: Should this be `enum` or `str`?

pkg/lang/frontend/starlark/io/io.go
```diff
@@ -15,18 +15,18 @@
 package io
 
 import (
-	"errors"
 	"os/user"
 	"path/filepath"
 	"strings"
 
+	"github.com/cockroachdb/errors"
+
```
gaocegege, line 23: Please merge these two groups. (Remove the new line here)

pkg/data/envd.go
VoVAllen: This is for starlark to show the type information of `data.envd`. Later I think we might have `data.s3`, `data.http`, etc., which have different type name I think. Do you think we should make it as enum?

pkg/home/data.go
VoVAllen: This is related to the uid/gid problem. 700 made it not accessible inside the container. I agree 777 is too much, I tried 644 before but doesn't work. I'll take a look on this problem again

pkg/data/envd.go
kemingy: I see. I thought we need to use pattern match later. If we only need to show the information, it's ok to be a string.

gaocegege:
PTAL @kemingy

gaocegege (approved):

kemingy (approved):

PR #757, opened 2022-08-11
fix: use defined jupyter port

kemingy:

* not specified: generate a free port
* specified: use it (even if it's not free) (just let it crash)

### DIFF
--- a/pkg/lang/ir/editor.go
+++ b/pkg/lang/ir/editor.go
@@ -20,6 +20,7 @@ import (
 	"github.com/cockroachdb/errors"
 	"github.com/moby/buildkit/client/llb"
 
+	"github.com/tensorchord/envd/pkg/config"
 	"github.com/tensorchord/envd/pkg/editor/vscode"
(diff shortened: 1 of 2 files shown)


gaocegege (approved):

gaocegege:
Maybe we should remove the JupyterConfig.Port directly in the future.

pkg/lang/ir/editor.go
```diff
@@ -85,7 +86,7 @@ func (g Graph) generateJupyterCommand(workingDir string) []string {
 			"--NotebookApp.token", "''")
 	}
 	if g.JupyterConfig.Port != 0 {
```
hezhizhen, line 88: The `if` statement is not necessary.

gaocegege: I think so, please help fix this if you are interested.

hezhizhen: https://github.com/tensorchord/envd/pull/773

PR #761, opened 2022-08-11
fix: set jupyter token with password config produced by build.envd

Xiaoaier-Z-L:
Signed-off-by: zhenguo.li <zhenguo.li@shopee.com>

### DIFF
--- a/pkg/lang/ir/editor.go
+++ b/pkg/lang/ir/editor.go
@@ -78,7 +78,7 @@ func (g Graph) generateJupyterCommand(workingDir string) []string {
 	}
 
 	if g.JupyterConfig.Password != "" {
-		cmd = append(cmd, "--NotebookApp.password", "''",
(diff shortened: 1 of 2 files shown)


pkg/lang/ir/editor.go
```diff
@@ -79,7 +79,7 @@ func (g Graph) generateJupyterCommand(workingDir string) []string {
 
 	if g.JupyterConfig.Password != "" {
 		cmd = append(cmd, "--NotebookApp.password", g.JupyterConfig.Password,
```
gaocegege, line 81: Should we set the password here?

gaocegege:
Please sign off both commits

pkg/lang/ir/editor.go
Xiaoaier-Z-L: Oh i want to rebase the 2 commits to 1 commits.

Issue #760, opened 2022-08-11
bug:  the jupyter password is invalid


Xiaoaier-Z-L:
## Description
i config the jupyter password by "test123456", but i can not enter it by this password.
when i check the log by "docker logs $containerName", it return 401 err
![image](https://user-images.githubusercontent.com/96805673/184114489-4adaeca4-72b9-4b02-8e11-bda17fa2582e.png)

## Reproduction
my build.envd config file
![image](https://user-images.githubusercontent.com/96805673/184114959-0514407e-980e-442f-8667-072d5f4a6d92.png)

## Additional Info
the envd main branch version

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritise the issues with the most 👍.

PR #762, opened 2022-08-11
fix: modify jupyter's authority from hash password to token string

Xiaoaier-Z-L:
Signed-off-by: zhenguo.li <zhenguo.li@shopee.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,7 +201,7 @@ def build():
         "numpy",
     ])
     shell("zsh")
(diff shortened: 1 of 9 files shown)


Xiaoaier-Z-L:
The jupyter password must be hase string, so i think we can set string token for jupyter authority

gaocegege:
Close https://github.com/tensorchord/envd/issues/760

PR #763, opened 2022-08-11
fix(CLI): 🔨  use latest version for local build

zwpaper:
1. detect version git tag for base image, set to `latest` if empty
2. update built.sh to add a tag env var, so that CI could use to change the tag to latest
3. update CI release to build a latest image when released, rebuilt one directly because there should be docker built cache to be used automatically.

Signed-off-by: Wei Zhang <kweizh@gmail.com>

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -116,6 +116,7 @@ jobs:
       run: |
           docker login --username "${DOCKERIO_USERNAME}" --password "${DOCKERIO_PASSWORD}"
           ./base-images/build.sh
+          DOCKER_IMAGE_TAG=latest ./base-images/build.sh
   cache_publish:
     name: Build & Push the remote cache
     # only trigger on main repo when tag starts with v
--- a/base-images/build.sh
+++ b/base-images/build.sh
@@ -13,9 +13,6 @@ ENVD_OS="${ENVD_OS:-ubuntu20.04}"
 JULIA_VERSION="${JULIA_VERSION:-1.8rc1}"
 RLANG_VERSION="${RLANG_VERSION:-4.2}"
 
-echo $ENVD_VERSION
-exit 1
-
 cd ${ROOT_DIR}
 # ubuntu 22.04 build require moby/buildkit version greater than 0.8.1
 if ! docker buildx inspect cuda; then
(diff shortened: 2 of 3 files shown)


base-images/build.sh
```diff
@@ -6,12 +6,15 @@ ROOT_DIR=`dirname $0`
...
 ENVD_VERSION="${ENVD_VERSION:-$GIT_TAG_VERSION}"
+DOCKER_IMAGE_TAG="${DOCKER_IMAGE_TAG:-$ENVD_VERSION}"
 DOCKER_HUB_ORG="${DOCKER_HUB_ORG:-tensorchord}"
 PYTHON_VERSION="${PYTHON_VERSION:-3.9}"
 ENVD_OS="${ENVD_OS:-ubuntu20.04}"
 JULIA_VERSION="${JULIA_VERSION:-1.8rc1}"
 RLANG_VERSION="${RLANG_VERSION:-4.2}"
 
+echo $ENVD_VERSION
+exit 1
```
gaocegege, line 17: Is this for debug purposes?

.github/workflows/release.yml
```diff
@@ -116,6 +116,7 @@ jobs:
       run: |
           docker login --username "${DOCKERIO_USERNAME}" --password "${DOCKERIO_PASSWORD}"
           ./base-images/build.sh
+          DOCKER_IMAGE_TAG=latest ./base-images/build.sh
```
gaocegege, line 119: What will happen if we release a new version? Will we still use latest as the tag?

gaocegege (commented):

Some comments here.

base-images/build.sh
zwpaper: haha, really sorry for this debug mistake, it must be too late yesterday night and I have been fall asleep 😭

.github/workflows/release.yml
zwpaper: yes, latest should always track on the newest release

base-images/build.sh
gaocegege: :smile: Please update the PR, and, take care.

zwpaper: It is a sad story, I tried to adjust my circadian clock, and found out I was a night cat at last🤣

PR updated

gaocegege (commented):
LGTM :+1:
/lgtm

BTW I like the PR title and commit message. The rise of the emoji!

PR #766, opened 2022-08-12
fix: -path and -file bug

### LINKED ISSUE #446: bug(CLI): fail to use both path and file flag in envd up

kemingy:

* fix #446

### DIFF
--- a/pkg/util/fileutil/file.go
+++ b/pkg/util/fileutil/file.go
@@ -51,6 +51,29 @@ func FileExists(filename string) (bool, error) {
 	return !info.IsDir(), nil
 }
 
+// FindFileAbsPath returns the absolute path for the given path and file
+func FindFileAbsPath(path, fileName string) (string, error) {
+	manifest := filepath.Join(path, fileName)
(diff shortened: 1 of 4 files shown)


pkg/util/fileutil/file.go
```diff
@@ -51,6 +51,29 @@ func FileExists(filename string) (bool, error) {
 	return !info.IsDir(), nil
 }
 
+// FindFileAbsPath returns the absolute path for the given path and file
+func FindFileAbsPath(path, fileName string) (string, error) {
```
gaocegege, line 55: Could we add a unit test for the func? Thanks!

kemingy: Sure

kemingy: Added. PTAL

gaocegege (approved):

PR #772, opened 2022-08-13
fix: Remove empty token arg

gaocegege:
There is no need to add `token=""` in jupyter func.

### DIFF
--- a/examples/mnist/build.envd
+++ b/examples/mnist/build.envd
@@ -9,7 +9,7 @@ def build():
         "numpy",
     ])
     shell("zsh")
-    config.jupyter(token="", port=8888)
+    config.jupyter()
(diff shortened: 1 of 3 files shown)


examples/mnist/build.envd
```diff
@@ -9,7 +9,7 @@ def build():
         "numpy",
     ])
     shell("zsh")
-    config.jupyter(token="", port=8888)
```
hezhizhen, line 12: default port?

gaocegege: The port is not needed, we will use a random port in the host to expose it.

kemingy (approved):

VoVAllen (approved):

PR #797, opened 2022-08-18
fix: add missing expose func exposed port to oci manifest

aseaday:


pkg/lang/ir/types.go
```diff
@@ -50,22 +50,28 @@ type Graph struct {
...
+
+	RuntimeGraph
+}
+
+// The results during runtime should be maintained here
+type RuntimeGraph struct {
+	RuntimeCommands map[string]string
+	RuntimeDaemon   [][]string
+	RuntimeEnviron  map[string]string
+	RuntimeExpose   *ExposeConfig
```
gaocegege, line 74: Can we remove the abstraction?
```suggestion
	ExposedPorts []ExposeItem
```

aseaday: > Can we remove the abstraction?

LGTM

gaocegege (approved):

PR #798, opened 2022-08-19
fix: set to latest if git tag is empty

hezhizhen:

### DIFF
--- a/pkg/lang/ir/compile.go
+++ b/pkg/lang/ir/compile.go
@@ -176,7 +176,7 @@ func (g Graph) DefaultCacheImporter() (*string, error) {
 	res := fmt.Sprintf(
 		"type=registry,ref=docker.io/%s/python-cache:envd-%s",
 		viper.GetString(flag.FlagDockerOrganization),
-		version.GetGitTagFromVersion())
+		version.GetVersionForImageTag())
 	return &res, nil
 }
 
--- a/pkg/lang/ir/compile.go
+++ b/pkg/lang/ir/compile.go
@@ -173,14 +173,10 @@ func (g Graph) EnvString() []string {
 
 func (g Graph) DefaultCacheImporter() (*string, error) {
 	// The base remote cache should work for all languages.
-	v := version.GetGitTagFromVersion()
-	if v == "" {
-		v = "latest"
-	}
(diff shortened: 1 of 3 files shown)


zwpaper:
unifying the duplicated codes in one place is a good idea, but I have a concern about the function `GetGitTagFromVersion()`, it's about `gitTag` maybe it should not return the docker image domain `latest`?

pkg/lang/ir/compile.go
```diff
@@ -173,14 +173,10 @@ func (g Graph) EnvString() []string {
...
 	// The base remote cache should work for all languages.
-	v := version.GetGitTagFromVersion()
-	if v == "" {
-		v = "latest"
-	}
 	res := fmt.Sprintf(
 		"type=registry,ref=docker.io/%s/python-cache:envd-%s",
 		viper.GetString(flag.FlagDockerOrganization),
-		v)
+		version.GetGitTagFromVersion())
```
gaocegege, line 179: Maybe we can rename the function.

hezhizhen: @zwpaper Any advice?

Thrimbda: IMHO, now it is this function's responsibility to retrieve the docker image's tag, then it may be appropriate to make that clearer from the name

hezhizhen: @Thrimbda Yes, and the question is which name will be better.
`GetImageTagFromVersion`?

aseaday: GetGitTag seems a internal implementation to define tag because the version is injected during building phase when `make` reads the git tag as the version. For a external use API, I think we shoud care about what the API is for finally.
It seems that this function give a short string as the version of a image tag. I think `GetVersionForImageTag` will be better.

zwpaper: `GetVersionForImageTag` seems great to me

gaocegege: GetVersionForImageTag LGTM

hezhizhen: PTAL

gaocegege (approved):

gaocegege:

Issue #782, opened 2022-08-16
feat: show expose service with port and name in envs subcommand


aseaday:
## Description

The expose services can be displayed via envd commands for users to know them.

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

aseaday:
envd envs describe --env envd-quick-start

```
DEPENDENCY      TYPE
...,
....
SERVICENAME PORT HOSTPORT
nginx                80     8080
```

gaocegege:
Looks promising.

aseaday:
@kemingy 圆梦大使感恩

PR #801, opened 2022-08-22
feat(CLI): envd env describe expose info

kemingy:

* close https://github.com/tensorchord/envd/issues/782

### DIFF
--- a/pkg/app/image_describe.go
+++ b/pkg/app/image_describe.go
@@ -15,6 +15,8 @@
 package app
 
 import (
+	"os"
+
 	"github.com/cockroachdb/errors"
 	"github.com/urfave/cli/v2"
 
@@ -48,6 +50,6 @@ func getImageDependency(clicontext *cli.Context) error {
 	if err != nil {
 		return errors.Wrap(err, "failed to list dependencies")
 	}
-	renderDependencies(dep)
+	renderDependencies(os.Stdout, dep)
 	return nil
 }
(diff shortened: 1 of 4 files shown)


kemingy:
image

* default sshd port is also displayed

pkg/app/image_describe.go
```diff
@@ -50,6 +48,6 @@ func getImageDependency(clicontext *cli.Context) error {
 	if err != nil {
 		return errors.Wrap(err, "failed to list dependencies")
 	}
-	renderDependencies(dep, os.Stdout)
```
gaocegege, line 53: It is for test purposes, should we keep it?

Or it is hard to test the feature in the future.

kemingy: Do you mean writing to `stdout` is for test purposes?

Even now, testing this feature is problematic unless we support output JSON.

aseaday: I guess gaoce's meaning is that we shoud pass the fd currently?

gaocegege: We may write UT for `renderDependencies`

kemingy: Fixed. PTAL.

aseaday: LGTM

aseaday (approved):
LGTM

gaocegege (approved):
LGTM :+1:
/lgtm

Issue #802, opened 2022-08-22
bug: e2e up test will download numpy even it's not required in `e2e/testdata/up-test`


kemingy:
## Description

* e2e up test will download `numpy` even it's not required in `e2e/testdata/up-test`
* because this test will use `https://pypi.tuna.tsinghua.edu.cn/` mirror, large package like `numpy` will take very long to download, and sometimes it will timeout, see https://github.com/tensorchord/envd/runs/7947345141?check_suite_focus=true#step:7:2452

## Additional Info

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritise the issues with the most 👍.

aseaday:
Let me have a look

VoVAllen:
@aseaday Probably related to the clean up problem like https://github.com/tensorchord/envd/blob/main/e2e/up_test.go#L68-L71

PR #803, opened 2022-08-22
chore: fix e2e cleanup by introduce new helper function ResetEnvdApp

aseaday:
https://github.com/tensorchord/envd/issues/802

### DIFF
--- a/e2e/quick_start_test.go
+++ b/e2e/quick_start_test.go
@@ -17,6 +17,7 @@ package e2e
 import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"github.com/tensorchord/envd/pkg/lang/ir"
 )
 
 var _ = Describe("e2e quickstart", Ordered, func() {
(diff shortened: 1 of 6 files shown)


aseaday:
I think it was caused by quick-start test, But `ir.DefaultGraph=new Graph` seems a little wired and ugly. Could we have a better way?

kemingy:
I tried to add it to https://github.com/tensorchord/envd/blob/5c4971b2f5fa6b32c2c247e18cf6c6a178d28f57/pkg/lang/ir/compile.go#L88-L90 but sometimes it doesn't work as expected.

kemingy:
BTW, you can search "pip install" in the e2e test log. You will find "pip install numpy numpy numpy".

aseaday:
> BTW, you can search "pip install" in the e2e test log. You will find "pip install numpy numpy numpy".

Yeah, That what I want to find out a clean way rather than dirty work one by one to fix it. Let me check the interpret logic because I refact it last week

e2e/quick_start_test.go
```diff
@@ -17,6 +17,7 @@ package e2e
 import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
```
gaocegege, line 19: Please add a blank line below.

aseaday: Thx and got it. Wait me to work out a clean way than I fix it.

gaocegege: Sure, take your time.

gaocegege (approved):
/lgtm
/approve

gaocegege (approved):

PR #804, opened 2022-08-23
test: add e2e test for examples in documentations

VoVAllen:


e2e/docs/testdata/ minimal/build.envd
```diff
@@ -0,0 +1,5 @@
+def build():
```
terrytangyuan, line 1: There's a leading space in the folder name `minimal`.

e2e/docs/docs_test.go
```diff
@@ -0,0 +1,83 @@
...
+		Expect(err).NotTo(HaveOccurred())
+
+		destroyArgs := append(baseArgs, []string{
+			"destroy", "--path", buildContext,
+		}...)
+		err = envdApp.Run(destroyArgs)
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	up_tests := []string{"testdata/minimal", "testdata/getting_started", "testdata/jupyter", "testsdata/complex"}
```
terrytangyuan, line 65: Should we move everything to `examples` and then test things in that folder directly? Otherwise, we need to maintain the copies in this folder.

VoVAllen: It's for the examples on the documentation site. Therefore no duplication here I think

e2e/docs/docs_test.go
```diff
@@ -0,0 +1,90 @@
...
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("can check envd envs details", func() {
+		buildContext := "testdata/minimal"
+		args := append(baseArgs, []string{
+			"up", "--path", buildContext, "--detach", "--force",
+		}...)
+		envdApp := app.New()
+		err := envdApp.Run(args)
```
gaocegege, line 54: We could add cleanup instruction in #803

e2e/docs/docs_test.go
kemingy: I wonder if we can insert these `envd` files into the doc instead of copying them here.

e2e/testdata/quick-start/build.envd
```diff
@@ -4,3 +4,5 @@ def build():
         "numpy",
     ])
     shell("zsh")
+
```
kemingy, line 7: I guess these two blank lines are generated by https://github.com/tensorchord/envd/blob/9e35cb8b4ab30e1aee638b88df3ef579a0154bd6/e2e/bytecode_hash_test.go#L25

Maybe we can find a better way later. So running e2e test locally won't modify the test files.

e2e/docs/docs_test.go
gaocegege: It would be great but may be hard to implement. I think copy/paste works in the current design.

gaocegege:
Please ping me if it is ready to review.

e2e/docs/docs_test.go
VoVAllen: https://vitepress.vuejs.org/guide/markdown#import-code-snippets
Vitepress support this, but bit hard since they're not in the same repo

VoVAllen:
@gaocegege PTAL

kemingy (approved):

e2e/docs/docs_test.go
kemingy: Will open a new issue to track this.

PR #815, opened 2022-08-26
WIP feat: add proposal for runtime graph image label

aseaday:


docs/proposals/20220826-image-runtime-redistributed-metadata.md
```diff
@@ -0,0 +1,23 @@
...
+the concept of `running context` is a also needed as addition of `build contxt` for example:
+
+- An engineer build a easy-to-use env for his/her interns for the quick use of company tools.
+- Kubernetes remote runtime support in the future.
+
+## Goals
+- Provide internal API:
+    - read runtime metadata
+    - write runtime metadata
+- `up` and `run` command doesn't need to interpret the `build.env`. `run` command can run a built env image with attachment of current working directory as `running context`.
```
gaocegege, line 22: The motivation LGTM. We need more implementation details. Looking forward to it!

aseaday: I am working on the implementation. Push the proposal first for others know what I am doing now.

docs/proposals/20220826-image-runtime-redistributed-metadata.md
```diff
@@ -0,0 +1,60 @@
...
+we use the following labels:
+
+- ai.tensorchord.envd.runtimeGraph.version
+- ai.tensorchord.envd.runtimeGraph.Daemon
+- ai.tensorchord.envd.runtimeGraph.Expose
+
+
+#### Why msgpack
+
+MessagePack is supported by over 50 programming languages and environments. And it will be shorter than JSON.
```
kemingy, line 60: Is it for `Result.Metadata` in buildkit gateway?

aseaday: It is for running phase. When you build a envd image and you want  run it without the manifest which contains information about port binding (such as `expose` keyword).

pkg/lang/ir/util.go
```diff
@@ -67,3 +69,29 @@ func getUIDGID() (int, int, error) {
 	}
 	return uid, gid, nil
 }
+
+func (rg *RuntimeGraph) DumpRuntimeGraph() (string, error) {
```
zwpaper, line 73: how about just `Dump`

pkg/lang/ir/util.go
```diff
@@ -67,3 +69,29 @@ func getUIDGID() (int, int, error) {
...
+	b, err := msgpack.Marshal(rg)
+	if err != nil {
+		return "", nil
+	}
+	runtimeGraphCode := b64.StdEncoding.EncodeToString(b)
+	return runtimeGraphCode, nil
+}
+
+func (rg *RuntimeGraph) LoadRuntimeGraph(code string) error {
```
zwpaper, line 82: how about just Load

go.mod
```diff
@@ -41,6 +41,8 @@ require (
 require (
 	github.com/op/go-logging v0.0.0-20160211212156-b2cb9fa56473 // indirect
 	github.com/pkg/errors v0.9.1 // indirect
+	github.com/vmihailenco/msgpack/v5 v5.3.5 // indirect
```
zwpaper, line 44: may I ask why we choose msgpack over json? as msgpack seems not that human-readable.

sometimes users may inspect an image to check the metadata, maybe it would be more convenient if the label is human-readable

zwpaper (changes requested):

pkg/lang/ir/util.go
aseaday: LGTM

pkg/lang/ir/util.go
aseaday: LGTM

go.mod
kemingy: Not sure if msgpack with base64 still has some advantage over JSON?

aseaday: Good question, let me test more cases.

aseaday: > may I ask why we choose msgpack over json? as msgpack seems not that human-readable.
> 
> sometimes users may inspect an image to check the metadata, maybe it would be more convenient if the label is human-readable

I think I could add a group of annotations for human readable purpose.

gaocegege: Same question. I prefer JSON here.

aseaday: Move to json now , PTAL @zwpaper @kemingy @gaocegege

kemingy (dismissed):

go.mod
gaocegege, line 44: I am wondering if the package is used. Should we rerun `go mod tidy`?

VoVAllen:
The only concern I have is whether there's any limit on the tag value length. However, I searched around and didn't find any explicitly stated limit, on docker hub/harbor/AWS ECR. I'm fine with this PR but probably we need some manual tests to see whether there's any potential compatibility issue. (i.e. for AWS ECR, the image tag may also be converted to AWS tag, which has a limit of 256 characters)

gaocegege (approved):

aseaday:
> The only concern I have is whether there's any limit on the tag value length. However, I searched around and didn't find any explicitly stated limit, on docker hub/harbor/AWS ECR. I'm fine with this PR but probably we need some manual tests to see whether there's any potential compatibility issue. (i.e. for AWS ECR, the image tag may also be converted to AWS tag, which has a limit of 256 characters)

I also don't find it in OCI spec but I don't know aws ecr has this limit. Let me give a cut off if length is over 256.

gaocegege:
> > The only concern I have is whether there's any limit on the tag value length. However, I searched around and didn't find any explicitly stated limit, on docker hub/harbor/AWS ECR. I'm fine with this PR but probably we need some manual tests to see whether there's any potential compatibility issue. (i.e. for AWS ECR, the image tag may also be converted to AWS tag, which has a limit of 256 characters)
> 
> I also don't find it in OCI spec but I don't know aws ecr has this limit. Let me give a cut off if length is over 256.

I do not think we should support such case since it is not the requirements of OCI spec.

aseaday:
> > > The only concern I have is whether there's any limit on the tag value length. However, I searched around and didn't find any explicitly stated limit, on docker hub/harbor/AWS ECR. I'm fine with this PR but probably we need some manual tests to see whether there's any potential compatibility issue. (i.e. for AWS ECR, the image tag may also be converted to AWS tag, which has a limit of 256 characters)
> > 
> > 
> > I also don't find it in OCI spec but I don't know aws ecr has this limit. Let me give a cut off if length is over 256.
> 
> I do not think we should support such case since it is not the requirements of OCI spec.

I think we can merge this PR now and open new issue to track runtime label. I am going to work on other issues now.

PR #831, opened 2022-08-30
bug: fix channels when use conda install with yaml file

VoVAllen:


pkg/lang/ir/parser/conda.go
```diff
@@ -24,6 +24,7 @@ type CondaEnv struct {
 	EnvName       string
 	CondaPackages []string
 	PipPackages   []string
+	Channels      []string
```
gaocegege, line 27: Could you please add a e2e test case in e2e/lang?

pkg/lang/ir/parser/conda.go
```diff
@@ -24,6 +24,7 @@ type CondaEnv struct {
...
 }
 
 func ParseCondaEnvYaml(condaEnvFile string) (*CondaEnv, error) {
```
kemingy, line 30: I wonder if we should use the `env.yaml` directly instead of parsing it?

Previously we need to parse it because we don't have access to the `env.yaml` during building. But now we're able to mount the workspace during building.

gaocegege: env.yaml contains the field `name`, which specifies the env name. But we hard coded the name to envd, thus there may be conflicts.

gaocegege: env.yaml contains the field `name`, which specifies the env name. But we hard coded the name to envd, thus there may be conflicts.

pkg/lang/ir/parser/conda.go
VoVAllen: done

VoVAllen:
@gaocegege test added

gaocegege (commented):
/lgtm

PR #832, opened 2022-08-31
fix(lang): expose host port

kemingy:

cc @aseaday

### DIFF
--- a/pkg/lang/frontend/starlark/runtime/runtime.go
+++ b/pkg/lang/frontend/starlark/runtime/runtime.go
@@ -95,7 +95,7 @@ func ruleFuncExpose(thread *starlark.Thread, _ *starlark.Builtin,
 	args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
 	var (
 		envdPort    starlark.Int
-		hostPort    = starlark.MakeInt(0)
+		hostPort    = starlark.MakeInt(0) // 0 means envd can randomly choose a free port
 		serviceName = starlark.String("")
 	)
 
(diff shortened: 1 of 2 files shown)


aseaday (approved):
LGTM

pkg/lang/frontend/starlark/runtime/runtime.go
```diff
@@ -95,7 +95,7 @@ func ruleFuncExpose(thread *starlark.Thread, _ *starlark.Builtin,
 	args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
 	var (
 		envdPort    starlark.Int
-		hostPort    = starlark.MakeInt(0)
+		hostPort    = starlark.MakeInt(0) // 0 means envd can randomly choose a free port
```
hezhizhen, line 98: When does envd randomly choose a port?

kemingy: If `host_port` is not specified, it will use the default value 0.

hezhizhen: So 0 is the default value for the host port (if users do not set it) instead of an indicator of a randomly assigned value for the host port. Do I understand it correctly?

kemingy: * 0 is the default value
* if it's 0, envd will randomly choose a free port, user can check the host port with `envd env describe --env <name>` https://github.com/tensorchord/envd/blob/404de3101cece0497084412433cf877f66cf5ee2/pkg/docker/docker.go#L514-L519

PR #838, opened 2022-09-01
feat(lang): install.python_packages(local_wheels=[])

### LINKED ISSUE #810: feat(lang): Support custom wheel
## Description

Users have to do this to install custom wheel

```python
def install_wheel(name):
    io.copy(src=name, dest="/")
    run(commands=["pip install /"+name])
```

We should support syntax like this:

```python
def build():
    install.packages(custom_wheel=[
        "xxx.wheel"
    ])
```

---

kemingy:

* fix #810

### DIFF
--- a/pkg/lang/ir/python.go
+++ b/pkg/lang/ir/python.go
@@ -28,6 +28,7 @@ import (
 
 const (
 	pythonVersionDefault = "3.9"
+	containerWheelPath = "/tmp/envd_wheel/"
 )
 
 func (g Graph) getAppropriatePythonVersion() (string, error) {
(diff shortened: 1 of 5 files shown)


pkg/lang/ir/python.go
```diff
@@ -166,6 +167,24 @@ func (g Graph) compilePyPIPackages(root llb.State) llb.State {
 			llb.Local(flag.FlagBuildContext), llb.Readonly)
 		root = run.Root()
 	}
+
```
gaocegege, line 170: https://github.com/tensorchord/envd/pull/838/files#diff-e9007374abf264af40157329f8dd5adff0461ef389c98a8daa32eaf54b5db8adR119 also should be updated.

pkg/lang/ir/python.go
```diff
@@ -166,6 +167,24 @@ func (g Graph) compilePyPIPackages(root llb.State) llb.State {
...
+	if len(g.PythonWheels) > 0 {
+		for _, wheel := range g.PythonWheels {
+			root = root.File(llb.Copy(
```
gaocegege, line 173: We could mount the build context dir like here.

https://github.com/tensorchord/envd/pull/838/files#diff-e9007374abf264af40157329f8dd5adff0461ef389c98a8daa32eaf54b5db8adL165

gaocegege (approved):
/lgtm

PR #880, opened 2022-09-12
refact: add envd home path func

kemingy:

### DIFF
--- a/pkg/util/fileutil/file.go
+++ b/pkg/util/fileutil/file.go
@@ -183,5 +183,5 @@ func DownloadOrUpdateGitRepo(url string) (path string, err error) {
 }
 
 func EnvdHomeDir(path ...string) string {
-	return filepath.Join(append([]string{"/home", "envd"}, path...)...)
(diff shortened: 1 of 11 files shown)


pkg/util/fileutil/file.go
```diff
@@ -181,3 +181,7 @@ func DownloadOrUpdateGitRepo(url string) (path string, err error) {
 
 	return path, nil
 }
+
+func EnvdHomeDir(path ...string) string {
+	return filepath.Join(append([]string{"/home", "envd"}, path...)...)
```
zwpaper, line 186: if we split `/home` and `envd`, should we also keep `/` isolated like `[]string{"/", "home", "envd"}`?

or if we put `/home` together, maybe it is not necessary to split the `envd`, make it `[]string{"/home/envd"}`?

kemingy: SGTM. Changed to `[]string{"/", "home", "envd"}`

gaocegege (commented):
/lgtm

I love this PR. A good start to refactor our messy code.

PR #886, opened 2022-09-14
fix(CLI): Fix build output argument and huggingface integration

gaocegege:
`--output does not work, this PR is to fix it.


kemingy (approved):
/lgtm

PR #887, opened 2022-09-14
chore: optimized the cache mechanism

aseaday:
This PR is to solve the followings problems:
- More and more deps files such as requirements.txt
- Include syntax that import other files

### DIFF
--- a/pkg/builder/builder.go
+++ b/pkg/builder/builder.go
@@ -20,6 +20,9 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"reflect"
+	"regexp"
+	"strings"
 
 	"github.com/cockroachdb/errors"
 	"github.com/moby/buildkit/client"
@@ -366,9 +369,10 @@ func (b generalBuilder) build(ctx context.Context, pw progresswriter.Writer) err
 
 func (b generalBuilder) checkIfNeedBuild(ctx context.Context) bool {
 	depsFiles := []string{
-		b.PubKeyPath,
 		b.ConfigFilePath,
 	}
+	depsFiles = getDepsFiles(depsFiles)
+	fmt.Println(depsFiles)
 	isUpdated, err := b.checkDepsFileUpdate(ctx, b.Tag, b.ManifestFilePath, depsFiles)
 	if err != nil {
 		b.logger.Debugf("failed to check manifest update: %s", err)
@@ -378,6 +382,65 @@ func (b generalBuilder) checkIfNeedBuild(ctx context.Context) bool {
 		return false
 	}
 	return true
+}
+
+func getDepsFiles(deps []string) []string {
+	tHandle := reflect.TypeOf(*ir.DefaultGraph)
+	vHandle := reflect.ValueOf(*ir.DefaultGraph)
+	deps = searchFileInGraph(tHandle, vHandle, deps)
+	return deps
+}
+
+// Match all filed in ir.Graph with the given keyword
+func likeFileFiled(str string) bool {
+	nameKeyword := []string{
+		"File",
+		"Path",
+		"Wheels",
+	}
+	if len(nameKeyword) == 0 {
+		return true
+	}
+	re := regexp.MustCompile(strings.Join(nameKeyword, "|"))
+	return re.MatchString(str)
+}
+
+// search all files in Graph
+func searchFileInGraph(tHandle reflect.Type, vHandle reflect.Value, deps []string) []string {
+	for i := 0; i < vHandle.NumField(); i++ {
+		v := vHandle.Field(i)
+		if v.Type().Kind() == reflect.Struct {
+			t := v.Type()
+			deps = searchFileInGraph(t, v, deps)
+		} else if v.Type().Kind() == reflect.Ptr {
+			if v.Type().Elem().Kind() == reflect.Struct {
+				if v.Elem().CanAddr() {
+					t := v.Type().Elem()
+					deps = searchFileInGraph(t, v.Elem(), deps)
+					fmt.Println((deps))
+				}
+			}
+		} else {
+			t := tHandle.Field(i)
+			fieldName := t.Name
+			if likeFileFiled(fieldName) {
+				typeName := t.Type.String()
+				if v.Interface() != nil {
+					if typeName == "string" {
+						deps = append(deps, v.Interface().(string))
+					}
+					if typeName == "*string" {
+						deps = append(deps, *(v.Interface().(*string)))
+					}
+					if typeName == "[]string" {
+						filesList := v.Interface().([]string)
+						deps = append(deps, filesList...)
+					}
+				}
+			}
+		}
+	}
+	return deps
 }
 
 // nolint:unparam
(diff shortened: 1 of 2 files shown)


gaocegege:
Thanks! Feel free to ping me if it is ready to review.

aseaday:
1. Use the hash of `ir.Graph` rather than the starlark bytecode hash to distinguish a manifest because of `include` syntax that will import other envd manifest and the graph is always be the right person.
2. Search all fields in `ir.Graph` with keywords like *file* and *path*. It is a little tricky because I don't want to define a hook and add it into every `compileFunc` in lang pkg. It is also a little dangerous If other contributors do not know this convention. So I need your suggestions @gaocegege @zwpaper

pkg/builder/builder.go
```diff
@@ -354,9 +355,10 @@ func (b generalBuilder) build(ctx context.Context, pw progresswriter.Writer) err
 
 func (b generalBuilder) checkIfNeedBuild(ctx context.Context) bool {
 	depsFiles := []string{
-		b.PubKeyPath,
 		b.ConfigFilePath,
 	}
+	depsFiles = getDepsFiles(depsFiles)
+	fmt.Println(depsFiles)
```
gaocegege, line 361: Please remove the fmt.Print statement here.

pkg/builder/builder.go
```diff
@@ -368,6 +370,65 @@ func (b generalBuilder) checkIfNeedBuild(ctx context.Context) bool {
 	return true
 }
 
+func getDepsFiles(deps []string) []string {
```
gaocegege, line 373: Is it possible to add the unit test cases for this func?

pkg/lang/ir/util.go
```diff
@@ -95,3 +99,14 @@ func (rg *RuntimeGraph) Load(code []byte) error {
 	rg.RuntimeExpose = newrg.RuntimeExpose
 	return nil
 }
+
+func GetDefaultGraphHash() string {
```
gaocegege, line 103: Could you please add a comment about this func? I cannot get how it does.

gaocegege (approved):
LGTM generally with some nits

gaocegege:
> Search all fields in ir.Graph with keywords like file and path. It is a little tricky because I don't want to define a hook and add it into every compileFunc in lang pkg. It is also a little dangerous If other contributors do not know this convention. So I need your suggestions @gaocegege @zwpaper

I think we should document it in the code. Maybe add the comments in DefaultGraph.

pkg/builder/builder.go
aseaday: I am working on it.

pkg/lang/ir/util.go
aseaday: I just found that It need a little fix because of pointer in the structure, I will add the document When I fix this bug.

pkg/builder/builder.go
aseaday: OK, maybe I should fix my local go-lint, It is very strange.

VoVAllen:
Does this mean any field name inside DefaultGraph, with "File"/"Wheel"/"Path" will be considered as dependency and invalidate caches when they change?

aseaday:
> Does this mean any field name inside DefaultGraph, with "File"/"Wheel"/"Path" will be considered as dependency and invalidate caches when they change?

Bingo!

pkg/builder/builder.go
```diff
@@ -368,6 +370,65 @@ func (b generalBuilder) checkIfNeedBuild(ctx context.Context) bool {
...
+		v := vHandle.Field(i)
+		if v.Type().Kind() == reflect.Struct {
+			t := v.Type()
+			deps = searchFileInGraph(t, v, deps)
+		} else if v.Type().Kind() == reflect.Ptr {
+			if v.Type().Elem().Kind() == reflect.Struct {
+				if v.Elem().CanAddr() {
+					t := v.Type().Elem()
+					deps = searchFileInGraph(t, v.Elem(), deps)
+					fmt.Println((deps))
```
kemingy, line 406: Another debug print.

pkg/builder/builder.go
```diff
@@ -368,6 +370,65 @@ func (b generalBuilder) checkIfNeedBuild(ctx context.Context) bool {
...
+	tHandle := reflect.TypeOf(*ir.DefaultGraph)
+	vHandle := reflect.ValueOf(*ir.DefaultGraph)
+	deps = searchFileInGraph(tHandle, vHandle, deps)
+	return deps
+}
+
+// Match all filed in ir.Graph with the given keyword
+func likeFileFiled(str string) bool {
+	nameKeyword := []string{
+		"File",
```
kemingy, line 383: Will this also match `HTTPInfo.Filename`? That one is an alias for the downloaded file.

pkg/builder/builder.go
aseaday: > < ~

pkg/builder/builder.go
VoVAllen: Would it better using struct tag?

pkg/builder/builder.go
terrytangyuan, line 361: Debugging statement is fine for certain cases, but we should probably switch to structured logging. Filed an issue to discuss separately https://github.com/tensorchord/envd/issues/889

aseaday: > Debugging statement is fine for certain cases, but we should probably switch to structured logging. Filed an issue to discuss separately #889

Good start for team work on debug/logging

pkg/builder/builder.go
aseaday: > Will this also match `HTTPInfo.Filename`? That one is an alias for the downloaded file.

I think it matched, let me try it.

pkg/builder/builder.go
```diff
@@ -380,6 +380,64 @@ func (b generalBuilder) checkIfNeedBuild(ctx context.Context) bool {
...
+		"Path",
+		"Wheels",
+	}
+	if len(nameKeyword) == 0 {
+		return true
+	}
```
kemingy, line 399: Will `len(nameKeyword) == 0` always be false?

pkg/builder/builder.go
kemingy: I think this will match `HTTPInfo.Filename`. That's not a file on the host. So the function `checkDepsFileUpdate` will always return **True**. Is that an expected behavior?

aseaday:
> Maybe we should always rebuild if there is a file from HTTP get.

LGTM

gaocegege:
Is this ready to review?

pkg/builder/builder.go
```diff
@@ -368,10 +368,13 @@ func (b generalBuilder) build(ctx context.Context, pw progresswriter.Writer) err
 }
 
 func (b generalBuilder) checkIfNeedBuild(ctx context.Context) bool {
+	if ir.DefaultGraph.HTTP != nil || ir.DefaultGraph.Repo != (types.RepoInfo{}) {
```
kemingy, line 371: `Repo` will only add an image label. `envd-server` will use a init container to pull and mount to the container.

So I guess this should not trigger a rebuild.

gaocegege: Same opinion here.

kemingy (approved):

aseaday:
@gaocegege @VoVAllen PTAL

gaocegege (approved):

PR #891, opened 2022-09-15
feat: support micromamba as an alternative to miniconda

kemingy:
Signed-off-by: Keming kemingyang@tensorchord.ai

Add the following code to your `~/.config/envd/config.envd` to try this new feature:

```py
config.conda_channel(use_mamba=True)
```

### DIFF
--- a/pkg/lang/ir/install-mamba.sh
+++ b/pkg/lang/ir/install-mamba.sh
@@ -11,4 +11,4 @@ mkdir -p ${MAMBA_BIN_DIR} && \
 curl -Ls https://micro.mamba.pm/api/micromamba/linux-${ARCH}/latest | tar -xvj -C ${MAMBA_BIN_DIR} --strip-components=1 bin/micromamba && \
 ln -s ${MAMBA_BIN_DIR}/micromamba ${MAMBA_BIN_DIR}/conda && \
 echo -e "channels:\n  - conda-forge" > ${MAMBA_ROOT_PREFIX}/.mambarc
-echo -e "#!/bin/sh\n\. ${MAMBA_ROOT_PREFIX}/etc/profile.d/micromamba.sh || return $?\nmicromamba activate \"$@\"" > ${MAMBA_BIN_DIR}/activate
+echo -e "#!/bin/sh\n\. ${MAMBA_ROOT_PREFIX}/etc/profile.d/micromamba.sh || return \$?\nmicromamba activate \"\$@\"" > ${MAMBA_BIN_DIR}/activate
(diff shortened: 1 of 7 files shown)


kemingy:
Benefit:

* smaller size (only a `micromamba` binary)
* no default Python environment
* faster than miniconda
* compatible with `conda` command (mostly)

pkg/lang/ir/install-mamba.sh
```diff
@@ -0,0 +1,14 @@
...
+    ARCH="aarch64"; \
+elif [[ "${ARCH}" == "ppc64le" ]]; then \
+    ARCH="ppc64le"; \
+else \
+    ARCH="64"; \
+fi && \
+mkdir -p ${MAMBA_BIN_DIR} && \
+curl -Ls https://micro.mamba.pm/api/micromamba/linux-${ARCH}/latest | tar -xvj -C ${MAMBA_BIN_DIR} --strip-components=1 bin/micromamba && \
+ln -s ${MAMBA_BIN_DIR}/micromamba ${MAMBA_BIN_DIR}/conda && \
+echo -e "channels:\n  - conda-forge" > ${MAMBA_ROOT_PREFIX}/.mambarc
```
kemingy, line 13: `conda-forge` is used by default.

This will install `numpy`, `scipy`, etc. with OpenBLAS. Check the doc here: https://numpy.org/install/#numpy-packages--accelerated-linear-algebra-libraries

pkg/lang/ir/install-mamba.sh
```diff
@@ -0,0 +1,14 @@
...
+echo -e "#!/bin/sh\n\. ${MAMBA_ROOT_PREFIX}/etc/profile.d/micromamba.sh || return \$?\nmicromamba activate \"\$@\"" > ${MAMBA_BIN_DIR}/activate
```
kemingy, line 14: Have to add an `activate` file like this to make sure `micromamba activate envd` works.

VoVAllen (commented):
Have you tested on both zsh & bash? 
Also Do you have any number of using mamba?

pkg/lang/ir/install-mamba.sh
VoVAllen: Does `condig.conda_channel` work for mamba?

kemingy:
> Have you tested on both zsh & bash? Also Do you have any number of using mamba?

Tested with the following setting:

```py
envdlib = include("https://github.com/tensorchord/envdlib")

def build():
    base(os="ubuntu20.04", language="python")
    config.conda_channel(use_mamba=True)
    envdlib.jupyter_lab(host_port=9000)
    shell("zsh") # shell(bash)
```

Not sure how many tests should we cover for mamba.

pkg/lang/ir/install-mamba.sh
kemingy: Yes. Check https://mamba.readthedocs.io/en/latest/user_guide/configuration.html#multiple-rc-files

kemingy:
Seems there are some incompatible commands for `update`.

* Check https://github.com/mamba-org/mamba/issues/1455

PR #929, opened 2022-09-21
chore: use conda-forge as default miniconda channel

kemingy:

According to:
* The founder of Anaconda https://www.reddit.com/r/Python/comments/iqsk3y/comment/gqdqrk9/
* Anaconda FAQ https://www.anaconda.com/blog/anaconda-commercial-edition-faq

I guess we should be able to use `miniconda` with the `conda-forge` channel.

BTW, we already support `micromamba` as an alternative. Although the CLI is not 100% compatible.


gaocegege (approved):
/lgtm

PR #932, opened 2022-09-21
feat: Support envd-server

gaocegege:
It is a significant change to envd. The PR is to:

- support `envd login`


pkg/app/ssh.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/cockroachdb/errors"
+	"github.com/urfave/cli/v2"
+
+	"github.com/tensorchord/envd/pkg/home"
+	"github.com/tensorchord/envd/pkg/ssh"
+	sshconfig "github.com/tensorchord/envd/pkg/ssh/config"
+)
+
+var CommandSSH = &cli.Command{
+	Name:     "ssh",
```
VoVAllen, line 27: Is this what we planned for `envd attach`? `attach` sounds better to me

VoVAllen (commented):
LGTM as the initial poc.

VoVAllen (approved):

pkg/app/ssh.go
gaocegege: I am not sure if we need this. the CLI UX design should be refined soon. Let's refine it later.

pkg/app/login.go
```diff
@@ -0,0 +1,79 @@
...
+		IdentityToken: resp.IdentityToken,
+	}, true); err != nil {
+		return errors.Wrap(err, "failed to create the auth config")
+	}
+	fmt.Println(resp.Status)
+	return nil
+}
+
+func GetMD5Hash(text string) string {
+	hasher := md5.New()
```
kemingy, line 76: Is MD5 secure in this scenario?

gaocegege: This token is used as the unique identifier of the user. Thus it should work.

kemingy (approved):

Issue #644, opened 2022-07-21
enhancement(build): configure user permissions takes too much time


gaocegege:
## Description

It takes too much time to configure the user permissions

```
=> [internal] create user group envd                                                                                                                                            0.4s
 => [internal] create user envd                                                                                                                                                  0.5s
 => [internal] add user envd to sudoers                                                                                                                                          0.4s
 => [internal] configure user permissions                                                                                                                                        0.3s
 => [internal] configure user permissions                                                                                                                                       60.8s
```

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

aseaday:
It is cause by  `chown -R`  
https://github.com/docker/for-linux/issues/388#issuecomment-1140409167 is a good idea but not for our case.

gaocegege:
Yep, I am not sure if we really need the -R chown since it is only to fix conda problems. Maybe we can have fine-grain control.

kemingy:
Close this because #933 fixed it.

PR #933, opened 2022-09-22
refact: conda/mamba create/update env, fix user permissions

kemingy:

* split "create user envd" and "chown envd"
* fix conda/mamba create new env from env YAML file


pkg/lang/ir/conda.go
```diff
@@ -66,97 +55,75 @@ func (g Graph) compileCondaChannel(root llb.State) llb.State {
...
+		env = fmt.Sprintf("--file %s ", g.CondaConfig.CondaEnvFileName) + env
+		// conda needs to use `conda env create` to create env from an env YAML file
+		if !g.CondaConfig.UseMicroMamba {
+			return fmt.Sprintf("%s env create %s", cmd, env)
+		}
+	}
+	return fmt.Sprintf("%s create %s", cmd, env)
+}
+
 func (g Graph) compileCondaPackages(root llb.State) llb.State {
```
gaocegege, line 86:
```suggestion
func (g *Graph) compileCondaPackages(root llb.State) llb.State {
```

VoVAllen (commented):
Do you need also chown on pip package path? Also if there's pip package in conda yaml file, is it installed under root user or envd user?

VoVAllen:
Also will conda install anything under current user folder?

kemingy:
> Do you need also chown on pip package path? Also if there's pip package in conda yaml file, is it installed under root user or envd user?

Pip package path should be under "/opt/conda/envs/envd".

Tested "install -e ." can work.

kemingy:
> Also will conda install anything under current user folder?

Does "current user folder" mean "/home/envd"? This one has also been chowned.

VoVAllen:
I mean if the installation process is executed under root user, will it install anything to `/root/XXXX`?

gaocegege:
> Pip package path should be under "/opt/conda/envs/envd".

Could you please verify if pip cache works after the change?

VoVAllen:
You can try install pip package such as numpy in conda. And check whether user can `import numpy` under user envd

gaocegege:
And, if we use root in the building, and use envd in the runtime, there may be different cache locations between build time and runtime. 

For example, pip cache in build time will be stored in /home/root/.pip, and it should be /home/envd/.pip in runtime.

I am not sure if it is possible to customise the cache location.

gaocegege:
Please rebase.

gaocegege:
https://github.com/tensorchord/envd/actions/runs/3170574497/jobs/5163236483#step:7:3192

There are some issues in CI.

kemingy:
According to my observation, `micromamba update -f env.yaml` doesn't install anything from the YAML file. Instead, it only updates all the packages.

We'd better use `conda update` because it requires mounts, which means it cannot be cached.

kemingy:
I removed most of the `chown` operations. Users can still access `conda`, use `pip install`.

* fix #644

kemingy:
`https://open-vsx.org/api` is too fragile. Sadly, GitHub actions don't support `allow-failure` https://github.com/actions/toolkit/issues/399.

gaocegege:
/lgtm

PR #959, opened 2022-09-30
bug: fix pypi package information

### LINKED ISSUE #937: bug: Wrong pip package information in `envd describe`
## Description

When installing package with `--extra-index-url` such as PyTorch, `--extra-index-url` will also be considered as a separate package

VoVAllen:

Fix #937


VoVAllen:
Open vscode plugsin seems in outage now

pkg/types/envd.go
```diff
@@ -240,12 +241,19 @@ func newDependencyFromLabels(label map[string]string) (*Dependency, error) {
 		}
```
gaocegege, line 241: Could we add the unit test for the func?

VoVAllen: added

gaocegege (approved):
LGTM :+1:
/lgtm

PR #962, opened 2022-10-01
fix(ir): `apt install` and `conda env create` cache

### LINKED ISSUE #897: bug: `conda create env` somtimes cannot be cached
## Description

Even though the specified Python version has not been changed, `envd` may not cache the `conda create env` process (not always). Which results in re-creating that may cost more than 30s.

This may be related to #666 

## Additional Info

---

### LINKED ISSUE #960: bug: remote cache for system packages doesn't work
## Description

For the 1st time build, cannot reuse the cache from tensorchord/python-cache for system packages.

## Additional Info

envd: v0.2.4-alpha.6

---

kemingy:

* fix #960 
* fix #897 

* Also, #933 is pending because of the same issue.


gaocegege (approved):
/lgtm

Issue #91, opened 2022-04-29
feasibility-research(lang): Refactor frontend language


gaocegege:
## Description

priority/high

gaocegege:
We may provide some primitives to play with buildkit in Go, then use these primitives to write logic in starlark. Like:

```python
def base(os, language):
    _base_img= Image("ubuntu:20.04")
    # Execute command over base
    install_htop = _base_img.run("sudo apt install htop")
    
    # as llb merge, create a new state as base the base image
    _base.image = _base.merge([install_htop, ...])

    # Later step will use _base.image as the base
```

gaocegege:
LLB Primitives we used in current ir package:

- `llb.Mkfile`
- `llb.User`
- `llb.Mkdir`
- `llb.WithCustomName`
- `llb.Shlex`
- `llb.Copy`
- `llb.Local`
- `llb.Merge`
- `llb.WithUIDGID`
- `llb.Diff`
- `llb.Scratch`
- `llb.Image`

LLB state operations:

- `state.Run`
- `state.AddMount`
-  `state.File`

gaocegege:
Some random thoughts about frontend language

```python
def build():
    base(os="ubuntu20.04", language="julia")
    
    config.julia_pkg_server(url="https://mirrors.tuna.tsinghua.edu.cn/julia")
    
    install.julia_packages([
        "Example"
    ])

    service.jupyter()
    service.ssh()
    service.new(name="tensorboard", command="tensorboard --logdir=logs")

    system.copy(src="example.ipynb", dst="notebooks/example.ipynb")
    
    vcs.git(remote="https://github.com/tensorchord/envd", branch="master", path="envd")

    data.dvc(remote="https://remote.dvc.org", path="data")
    data.s3(bucket="tensorchord-data", path="data")
```

kemingy:
The design of `state` in `bulidkit` looks suitable for method chaining.

*Pros*
* commands run in order
* users can control which parts are run in parallel

*Cons*
* need to refactor a lot, now we build a static graph instead of a DAG

gaocegege:
The interface will look like:

```
Image("ubuntu20.04").apt_install("make").with_python("3.9").with_cuda()
Scratch().git_repo("xxx").copy_to(img, src="/opt", dest="/")
```

PR #972, opened 2022-10-07
proposal(ir): state based implementation

kemingy:

* related to #91

### DIFF
--- a/docs/proposals/20221007-ir-state.md
+++ b/docs/proposals/20221007-ir-state.md
@@ -1,0 +1,53 @@
+# Refactor the IR to use state
+
+Author:
+- [kemingy](https://github.com/kemingy/)
+
+
+## Summary
+
+This proposal is designed to refactor the IR, as we discussed in [envd/#91](https://github.com/tensorchord/envd/issues/91#issuecomment-1239405359).
+
+## Motivation
+
+* Currently, we build the image with hard-coded order and parallelism. It's not flexible enough for some use cases.
+* The design of buildkit LLB state is suitable for method chaining implementation. Users can define the dependencies and parallelism easily.
+
+## Goals
+
+* Able to define the order and parallelism.
+* Provide low level APIs to build the image from scratch.
+
+## API
+
+```python
+def conda_env(root):
+    conda = root.env("MAMBA_ROOT_PREFIX", "/opt/conda").run(
+        [
+            "curl micro.mamba.pm/install.sh | sh -s -- -y",
+            "/opt/conda/bin/micromamba create -n envd python={}".format("3.10"),
+        ]
+    )
+    return conda
+
+
+def build():
+    root = (
+        Image("ubuntu:22.04")
+        .apt_packages(
+            ["curl", "wget"], update=True, clean=True, without_recommends=True
+        )
+        .run(["curl https://starship.rs/install.sh | sh -s -- -y"], mount=None)
+        .copy(
+            Image("tensorchord/envd-sshd-from-scratch:latest"),
+            source_path="/usr/bin/envd-sshd",
+            dest_path="/var/envd/bin/envd-sshd",
+        )
+    )
+    return conda_env(root)
+```
+
+## Risk
+
+* Implementation in Starlark
+* VSCode language server


VoVAllen:
How to merge different different stages like llb.Merge?

Also I think [bramble](https://github.com/maxmcd/bramble)'s syntax can be an option

kemingy:
> How to merge different different stages like llb.Merge?

Will add `Merge`, `Diff`, `File` later.

> Also I think [bramble](https://github.com/maxmcd/bramble)'s syntax can be an option

Will take a look.

VoVAllen:
Some examples from bramble https://github.com/maxmcd/bramble/blob/eea4aee51e6ad881166412d61190012fb0d97c56/internal/project/testdata/project/default.bramble

VoVAllen:
I think we should have an idea about what the llb graph will look like after more dependency is set by the user (such as gcc and pypi packages), that can also utilize caches as much as possible. 

>  Simple things should be simple, complex things should be possible.

gaocegege:
I think bramble's example there looks complex.

https://github.com/maxmcd/bramble/blob/eea4aee51e6ad881166412d61190012fb0d97c56/internal/project/testdata/project/default.bramble

It declares the input arguments explicitly. Personally, I prefer the func chain.

kemingy:
> I think we should have an idea about what the llb graph will look like after more dependency is set by the user (such as gcc and pypi packages), that can also utilize caches as much as possible.
> 
> > Simple things should be simple, complex things should be possible.

Agree. What we have now should be simple. Others like parallelism should be possible.

Method chaining should be enough for sequence commands. Each function should return a `state`. (`ExecState` should be hidden by introducing more parameters)

gaocegege:
Some questions:

- Is it possible to auto-merge the two chains? Merge/diff should be advanced statements.
- How to integrate it with the envdlib?

docs/proposals/20221007-ir-state.md
```diff
@@ -0,0 +1,81 @@
...
+
+
+def install_vscode_ext(ext=()):
+    root = Source.merge(Source.http("openvsx/{}".format(x), filename=x) for x in ext)
+    return root
+
+
+def install_key():
+    root = (
+        Source.scratch()
```
gaocegege, line 46: Could we remove the `Source`?

Use `scratch`, `merge` and `image`.

gaocegege:
Then could the new language syntax be compatible with the existing design?

Or is it a total breaking change?

BTW, could you please provide the example for `python-basic` with the new design?

docs/proposals/20221007-ir-state.md
kemingy: fixed

kemingy:
> Then could the new language syntax be compatible with the existing design?
> 
> Or is it a total breaking change?

I think it will be a breaking change.

> BTW, could you please provide the example for `python-basic` with the new design?

Already been added to the proposal. PTAL.

gaocegege:
@VoVAllen WDYT

I have no opinion on it, let's start researching if starlark supports it.

VoVAllen:
I'm a bit concerned about the current proposal. The current design makes original usage becomes detail-oriented, which is more complicated than original design.
Also current design looks similar to llb, we can also consider expose llb-like primitives directly. Explicit dependency declaration is an advanced function, llb primitives would be easier for us to maintain and ensures that "complex thing is possible"

Some personal thoughts:
Explicitly define two/three stages. base stages, envd-managed stages(`install.python_packages` etc.), user-managed stages(`run(XXX)`)

The difference between them is:
- base stages can be overwritten by custom images, and managed by envd if not specified
- envd-managed stages will parallelize and use cache as much as possible to accelerate the process, thus no dependency can be set here.
- user-managed stages can be fully customized, with explicit dependency.

Other ideas:

All functions provided by envd can add a new argument, such as called `state`. 

In user stages user can do:
```
state = stage('user')
state1 = install.apt_packages(["g++"], state=state)
new_state = install.python_packages(["package_needs_g++"], state=state1)
```
and to define it as a custom function:
```
def install_inhouse_package():
  state = stage('user')
  state1 = install.apt_packages(["g++"], state=state)
  new_state = install.python_packages(["package_needs_g++"], state=state1)
  # envd_output is an builtin variable, add means merge state with the final output
  envd_output.add(new_state)
  return new_state
```
To use
```
def build():
   install.python_packages(['torch'])
   install_inhouse_packages()
```

WDYT

kemingy:
> I'm a bit concerned about the current proposal. The current design is detail-oriented, which is more complicated than original design. Also current design looks similar to llb, we can also consider expose llb-like primitives directly. Explicit dependency declaration is an advanced function, llb primitives would be easier for us to maintain and ensures that "complex thing is possible"
> 
> Some personal thoughts: Explicitly define two/three stages. base stages, envd-managed stages(`install.python_packages` etc.), user-managed stages(`run(XXX)`)
> 
> The difference between them is:
> 
>     * base stages can be overwritten by custom images, and managed by envd if not specified
> 
>     * envd-managed stages will parallelize and use cache as much as possible to accelerate the process, thus no dependency can be set here.
> 
>     * user-managed stages can be fully customized, with explicit dependency.

This is similar to the current implementation and this proposal. We do have different stages, it's just not explicit.

We can provide the `install.conda_python()` function. So users who start with the custom images can use it to install the python environment.

> Other ideas:
> 
> All functions provided by envd can add a new argument, such as called `state`.
> 
> In user stages user can do:
> 
> ```
> state = stage('user')
> state1 = install.apt_packages(["g++"], state=state)
> new_state = install.python_packages(["package_needs_g++"], state=state1)
> ```
> 
> and to define it as a custom function:
> 
> ```
> def install_inhouse_package():
>   state = stage('user')
>   state1 = install.apt_packages(["g++"], state=state)
>   new_state = install.python_packages(["package_needs_g++"], state=state1)
>   # envd_output is an builtin variable, add means merge state with the final output
>   envd_output.add(new_state)
>   return new_state
> ```
> 
> To use
> 
> ```
> def build():
>    install.python_packages(['torch'])
>    install_inhouse_packages()
> ```
> 
> WDYT

Defining the dependencies with an extra state argument is acceptable but not very user-friendly.

The LLB-like syntax is only complex when you need to use `diff` and `merge`. Otherwise, the method chaining should be a simple solution.

kemingy:
One more thing, this is incompatible with `config.envd`.

PR #982, opened 2022-10-08
fix: Remove dump checkout and remove pre-commit

gaocegege:

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -44,7 +44,6 @@ jobs:
         run: |
           make addlicense && git add pkg cmd &&
           git diff --cached --exit-code || (echo 'Please run "make addlicense" to verify govet' && exit 1);
-      - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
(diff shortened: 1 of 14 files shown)


gaocegege:
@hawkingrei 

noloopclosure is not merged in v1.50, thus we cannot use it now.

Ref https://github.com/golangci/golangci-lint/pull/3027

gaocegege:
I removed the pre-commit here because it is buggy. golang lint will not run if only .github/flows are changed. Thus we do not know if the change to golangci-lint works.

hawkingrei:
Of course， ```noloopclosure``` is a good linter. it has been enabled in the TiDB repo. it can be enabled when is merged into the golangci-linter.

gaocegege:
I will open an issue to keep track of it.

VoVAllen (approved):
LGTM. What's the difference comparing to the original lint?

gaocegege:
The golangci-lint action will be run every time.

The previous implementation only run when the file changed.

kemingy (approved):
\lgtm

Issue #892, opened 2022-09-15
feat(build): detect if the current environment is running before building


kemingy:
## Description

Currently, the logic is: build -> check if the same container is running -> remind the user to use `--force` or manually clean it.

We should make it to: detect the current environment -> remind the user if it's running -> build

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

Belyenochi:
If no one else, please assign it to me :-).

kemingy:
> If no one else, please assign it to me :-).

/assign @Belyenochi

gaocegege:
Thanks! 

And, ask us if there is any question about this feature.

Here is the discord invitation link: https://discord.gg/KqswhpVgdU

gaocegege:
Hi @Belyenochi 

Are you still working on the issue?

Belyenochi:
> Hi @Belyenochi
> 
> Are you still working on the issue?

Sorry for not updating this issue, I will try to submit this MR this week.

Belyenochi:
@gaocegege , hi gaocegege, I have two not so sure about the implementation of this problem
1. Whether the original image that uses --force to rebuild the image needs to be retained, I understand that it may be necessary
2. When reminding the user that the same container is running, whether the user needs to use a judgment like yes or no to allow or block the build or is it just a reminder and the default result is the same as the original --force logic

gaocegege:
@kemingy Could you please help answer the questions?

kemingy:
>     1. Whether the original mirroring logic of rebuilding the image through --force needs to be retained, I understand that it may be necessary

Yes. We should keep the `--force`.

>     2. When reminding the user that the same container is running, whether the user needs to use a judgment like yes or no to allow or block the build or is it just a reminder and the default result is the same as the original --force logic

I think a reminder should be easy to implement.

Belyenochi:
> > ```
> > 1. Whether the original mirroring logic of rebuilding the image through --force needs to be retained, I understand that it may be necessary
> > ```
> 
> Yes. We should keep the `--force`.
> 
> > ```
> > 2. When reminding the user that the same container is running, whether the user needs to use a judgment like yes or no to allow or block the build or is it just a reminder and the default result is the same as the original --force logic
> > ```
> 
> I think a reminder should be easy to implement.

thanks @kemingy @gaocegege  :-).

PR #989, opened 2022-10-09
feat(build): detect if the current environment is running before building (#892)

Belyenochi:
support [#892](https://github.com/tensorchord/envd/issues/892)

### DIFF
--- a/pkg/app/build.go
+++ b/pkg/app/build.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/tensorchord/envd/pkg/builder"
 	"github.com/tensorchord/envd/pkg/docker"
+	"github.com/tensorchord/envd/pkg/envd"
 	"github.com/tensorchord/envd/pkg/home"
 	sshconfig "github.com/tensorchord/envd/pkg/ssh/config"
 	"github.com/tensorchord/envd/pkg/util/fileutil"
@@ -124,13 +125,13 @@ func build(clicontext *cli.Context) error {
 }
 
 func DetectEnvironment(clicontext *cli.Context, buildOpt builder.Options) error {
-	dockerClient, err := docker.NewClient(clicontext.Context)
+	engine, err := envd.New(clicontext.Context, "docker")
 	if err != nil {
 		return errors.Wrap(err, "failed to create the docker client")
 	}
 	// detect if the current environment is running before building
 	ctr := filepath.Base(buildOpt.BuildContextDir)
-	running, _ := dockerClient.IsRunning(clicontext.Context, ctr)
+	running, _ := engine.IsRunning(clicontext.Context, ctr)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


gaocegege (commented):
LGTM

/cc @kemingy 

pkg/app/build.go
```diff
@@ -123,6 +124,23 @@ func build(clicontext *cli.Context) error {
...
+	engine, err := envd.New(clicontext.Context, "docker")
+	if err != nil {
+		return errors.Wrap(err, "failed to create the docker client")
+	}
+	// detect if the current environment is running before building
+	ctr := filepath.Base(buildOpt.BuildContextDir)
+	running, _ := engine.IsRunning(clicontext.Context, ctr)
+	if err != nil {
+		return err
+	}
```
kemingy, line 137: What is this `err`?

pkg/app/build.go
```diff
@@ -123,6 +124,23 @@ func build(clicontext *cli.Context) error {
...
+	if running {
+		logrus.Warnf("detect container %s is running.", ctr)
```
kemingy, line 139: I think it should return an error if `--force` is not set. Because it will fail later.

pkg/app/build.go
Belyenochi: sorry, my fault

pkg/app/build.go
Belyenochi: > I think it should return an error if `--force` is not set. Because it will fail later.
If so I think this part of the logic overlaps with the logic of func (e dockerEngine) CleanEnvdIfExists(ctx context.Context, name string, force bool) error I tend to remove the code in CleanEnvdIfExists.

Belyenochi: The modified code looks like this.
```
func (e dockerEngine) CleanEnvdIfExists(ctx context.Context, name string, force bool) error {
	created, err := e.Exists(ctx, name)
	if err != nil {
		return err
	}
	if !created {
		return nil
	}
	return e.ContainerRemove(ctx, name, dockertypes.ContainerRemoveOptions{Force: force})
}
```

kemingy: cc @gaocegege should we bring the `CleanEnvdIfExists` before `BuildImage`?

`CleanEnvdIfExists` and `DetectEnvironment` are doing something similar.

gaocegege: Personally, prefer removing the code in CleanEnvdIfExists.

The func `CleanEnvdIfExists` is defined in envd.Engine. And the engine interface will have two implementations soon. Thus it will be great to keep the logic there simple and clean.

WDYT @kemingy

Belyenochi: @kemingy  @gaocegege I tried to accomplish this modification.

kemingy (approved):
/lgtm

gaocegege:
/approve

Thanks for your contribution!

gaocegege:
@all-contributors Please add @Belyenochi for the code contribution.

Issue #101, opened 2022-05-05
feat(CLI): Support SSH(attach) command


gaocegege:
## Description

midi ssh

kemingy:
Is it done though sshd?

gaocegege:
This issue is to support a new command `envd ssh` to attach into the env.

kemingy:
> This issue is to support a new command `envd ssh` to attach into the env.

What's the main difference between `ssh <envd-container>` and `envd ssh <envd-container>`? Any feature we should support through this new command?

gaocegege:
envd ssh or envd attach does not require the target if it is run in the build context dir.

gaocegege:
For envd-server runner:

`envd ssh` will be introduced as the new command. It creates the ssh client config, then connect to the ssh proxy server `containerssh`. The username should be the contains the identitytoken and project name.

The `envd-server` should:

- Get the identity token and the project name, then find the correct pod to attach.
- Update `containerssh` app config with the `/v1/config` endpoint.

PR #992, opened 2022-10-09
feat: Support cli argument for host key

gaocegege:
> If you don't specify a host key, it will generate one every time. This is convenient except you'll have to deal with clients being confused that the host key is different. It's a better idea to generate or point to an existing key on your system:

https://pkg.go.dev/github.com/gliderlabs/ssh@v0.3.5

The host key in envd-sshd is generated every time. Thus the fingerprint of the sshd always changes. This PR is to add a cli argument to support custom host key.

Ref #101 

### DIFF
--- a/pkg/ssh/ssh.go
+++ b/pkg/ssh/ssh.go
@@ -134,7 +134,7 @@ func NewClient(opt Options) (Client, error) {
 			// create agent and add in auth
 			forwardingAgent := agent.NewClient(agentConn)
 			// add callback for forwarding agent to SSH config
-			// XXX - might want to handle reconnects appending multiple callbacks
+			// might want to handle reconnects appending multiple callbacks
 			auth := ssh.PublicKeysCallback(forwardingAgent.Signers)
 			config.Auth = append(config.Auth, auth)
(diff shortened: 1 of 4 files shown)


pkg/ssh/ssh.go
```diff
@@ -120,26 +123,31 @@ func NewClient(opt Options) (Client, error) {
...
+		socketLocation := os.Getenv("SSH_AUTH_SOCK")
+		if socketLocation != "" {
+			agentConn, err := net.Dial("unix", socketLocation)
+			if err != nil {
+				return nil, errors.Wrap(err, "could not connect to local agent socket")
+			}
+			// create agent and add in auth
+			forwardingAgent := agent.NewClient(agentConn)
+			// add callback for forwarding agent to SSH config
+			// XXX - might want to handle reconnects appending multiple callbacks
```
VoVAllen, line 137: What's XXX?

VoVAllen (commented):
What's host key? Is it public key? Or is it `flagAuthKey`?

gaocegege:
Ref https://www.ssh.com/academy/ssh/host-key

VoVAllen (approved):
LGTM.

PR #1007, opened 2022-10-11
feat(cli): add `envd images remove` command

### LINKED ISSUE #975: feat(CLI): Add envd images remove command
## Description

We lack the command `envd images remove` to remove the envd images. The environment can be removed via `envd destroy`, but the built images cannot be removed using `envd`. This issue is to support it.

---

Triple-Z:
resolves #975

### DIFF
--- a/pkg/app/image_remove.go
+++ b/pkg/app/image_remove.go
@@ -1,0 +1,68 @@
+// Copyright 2022 The envd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package app
+
+import (
+	"fmt"
+
+	"github.com/cockroachdb/errors"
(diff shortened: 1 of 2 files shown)


pkg/app/image_remove.go
```diff
@@ -0,0 +1,68 @@
...
+	Name:    "remove",
+	Aliases: []string{"r"},
+	Usage:   "Remove an envd image",
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:    "image",
+			Usage:   "Specify the image name to be removed",
+			Aliases: []string{"i"},
+		},
+		&cli.StringFlag{
```
gaocegege, line 37: Could we support --image image-name:tag?

Triple-Z: So the `images remove` command should only have one `--image` flag, and no need for `--tag` flag.

The possible usage like the following.
```
envd images remove --image name  # with default tag `dev`
envd images remove --image name:tag
```

gaocegege: The default tag should be `latest`, I think.

gaocegege:
```
envd images remove --image name  # with default tag `latest`
envd images remove --image name:tag
```

Triple-Z: But the default tag in `envd build` is `dev`. 

https://github.com/tensorchord/envd/blob/f26fc4ce7e42ed0e567cb492c41551f6c2f2c3b4/pkg/app/build.go#L197-L200

Triple-Z: > Could we support --image image-name:tag?

Supported, PTAL @gaocegege

gaocegege (approved):
/lgtm

PR #1008, opened 2022-10-11
bug: fix explicit channel setting with conda env.yaml

### LINKED ISSUE #1005: bug(conda): Fail to use conda channel when the env.yaml is given
## Description

Seems that `config.conda_channel` does not work with `install.conda_packages(env=xx)`

https://github.com/tensorchord/envd/blob/main/pkg/lang/ir/conda.go#L99

/cc @kemingy @VoVAllen 

## Additional Info

---

VoVAllen:

Fix #1005

### DIFF
--- a/e2e/language/testdata/python/conda_channel/build.envd
+++ b/e2e/language/testdata/python/conda_channel/build.envd
@@ -3,7 +3,7 @@ def build():
         channel="""
 channels:
   - defaults
-  - conda-forge # dgllife in env.yaml only works if specify channels here
+  - dglteam # dgllife in env.yaml only works if specify channels here
     """
     )
     install.conda_packages(env_file="env.yaml")
--- a/e2e/language/testdata/python/conda_channel/build.envd
+++ b/e2e/language/testdata/python/conda_channel/build.envd
@@ -4,6 +4,7 @@ def build():
 channels:
   - defaults
   - conda-forge # dgllife in env.yaml only works if specify channels here
+    """
     )
     install.conda_packages(env_file="env.yaml")
     base(os="ubuntu20.04", language="python3.8")
--- a/e2e/language/testdata/python/conda_channel/build.envd
+++ b/e2e/language/testdata/python/conda_channel/build.envd
@@ -1,0 +1,9 @@
+def build():
+    config.conda_channel(
+        channel="""
+channels:
+  - defaults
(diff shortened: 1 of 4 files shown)


pkg/lang/ir/conda.go
```diff
@@ -39,7 +39,7 @@ const (
 var (
 	// this file can be used by both conda and mamba
 	// https://mamba.readthedocs.io/en/latest/user_guide/configuration.html#multiple-rc-files
-	condarc = fileutil.EnvdHomeDir(".condarc")
+	condarc = "/opt/conda/.condarc"
```
gaocegege, line 42: lgtm

/cc @kemingy

kemingy: Will this override https://github.com/tensorchord/envd/blob/d87afda9213fcb334ca55d247ba3c0663b7b180b/pkg/lang/ir/install-conda.sh#L24

But I guess it's okay?

e2e/language/testdata/python/conda_channel/build.envd
```diff
@@ -0,0 +1,10 @@
+def build():
+    config.conda_channel(
+        channel="""
+channels:
+  - defaults
+  - conda-forge # dgllife in env.yaml only works if specify channels here
```
kemingy, line 6: https://github.com/tensorchord/envd/blob/d87afda9213fcb334ca55d247ba3c0663b7b180b/pkg/lang/ir/install-conda.sh#L24
Already use `conda-forge`. So not sure if this test case can fail.

VoVAllen: I think it's fine. The condarc in install-conda.sh is used to install python. And the conda_channel is for user configuration, I think it's fine to overwrite?

kemingy: It's okay to override. But this won't be able to detect if the new channel config is right or wrong. Because even if it doesn't work, the default channel is already `conda-froge`, and the test always passes.

VoVAllen: Changed to use dgl as test target

VoVAllen: Changed to use dgl which needs `dglteam` channel for testing

kemingy (approved):
/lgtm

PR #1024, opened 2022-10-13
bug: Fix install from source will lose version information

VoVAllen:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,7 +71,7 @@ BUILD_DIR := ./build
 VERSION ?= $(shell git describe --match 'v[0-9]*' --always --tags --abbrev=0)
 BUILD_DATE=$(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
 GIT_COMMIT=$(shell git rev-parse HEAD)
-GIT_TAG=$(shell if [ -z "`git status --porcelain`" ]; then git describe --exact-match --tags HEAD 2>/dev/null; fi)
+GIT_TAG ?= $(shell if [ -z "`git status --porcelain`" ]; then git describe --exact-match --tags HEAD 2>/dev/null; fi)
 GIT_TREE_STATE=$(shell if [ -z "`git status --porcelain`" ]; then echo "clean" ; else echo "dirty"; fi)
 GITSHA ?= $(shell git rev-parse --short HEAD)
 GIT_LATEST_TAG ?= $(shell git describe --tags --abbrev=0)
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -26,7 +26,15 @@
 def build_envd_if_not_found():
     if not os.path.exists("bin/envd"):
         logging.info("envd not found. Build from scratch")
-        errno = subprocess.call(["make", "build-release"])
+        try:
+            with open(".GIT_TAG_INFO") as f:
```
gaocegege, line 30: Where do we write the file? Is it maintained by us?

dragonly: https://github.com/tensorchord/envd/pull/1024/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R113
Seems like here. Newly added.

dragonly: However, I didn't get the whole logic. Can you explain the design and execution order? Thanks! @VoVAllen

setup.py
```diff
@@ -26,7 +26,15 @@
...
+                logging.info("Use build_tag from envd._version")
+                tag = f.read().strip()
+                errno = subprocess.call(
+                    ["GIT_TAG={}".format(tag), "make", "build-release"]
```
dragonly, line 34: It would be more explicit to use the `env=` parameter to pass the environment variable.
https://docs.python.org/3/library/subprocess.html#popen-constructor

setup.py
VoVAllen: On macos arm, it built from source without git info. But the whole logic related to version depends on git tag. Missing git tag results in the bug. Thus here we manually save git tag info here

Makefile
```diff
@@ -110,6 +110,7 @@ build-release:
 		-X $(ROOT)/pkg/version.gitTag=$(GIT_TAG)" \
 	    $(CMD_DIR)/$${target};                                                         \
 	done
+	@[[ ! -z "$(GIT_TAG)" ]] && echo "$(GIT_TAG)" > .GIT_TAG_INFO
```
kemingy, line 113: Will this file be packed into the source tar? I think it's better to implicitly add to the `MANIFEST.in`.

setup.py
```diff
@@ -26,7 +26,15 @@
...
+                )
+        except:
+            errno = subprocess.call(["make", "build-release", ""])
```
kemingy, line 37: What is the purpose of the last empty string?

setup.py
VoVAllen: fixed

setup.py
VoVAllen: fixed

Makefile
VoVAllen: added

VoVAllen:
Tested on my macos, now envd binary from source has a proper version

kemingy (approved):

setup.py
dragonly: nit: It's better to add comments here to explain the necessity of the code block. If one day macOS pre-built binary is released, this can be removed by consulting to the comment here.

dragonly (approved):
LGTM. Goot Job!

Issue #552, opened 2022-07-05
enhancement(CLI): zsh-completion warning


shaonianche:
When running the `envd bootstrap` command in a device without the `zsh shell` installed, the following Warning occurs：

```bash
WARN[2022-07-05T10:12:25Z] Warning: unable to enable zsh-completion: /usr/local/share/zsh/site-functions does not exist
```

gaocegege:
We should check if zsh exists. And remove the warning.

shaonianche:
> We should check if zsh exists. And remove the warning.

thx!

gaocegege:
It is not fixed, I just explain more about the issue. Let's reopen it.

dragonly:
I have installed zsh and OhMyZsh, but there's still this warning. OS is macOS.

```bash
> envd bootstrap
...
WARN[2022-10-11T21:34:37+08:00] Warning: unable to enable zsh-completion: /usr/local/share/zsh/site-functions does not exist
...
```

Then I manually created this directory, but still with no luck
```
WARN[2022-10-11T21:59:50+08:00] Warning: open /usr/local/share/zsh/site-functions/_envd: permission denied
```

Maybe it's better to search for $fpath and find out the proper directory to install the autocomplete scripts?
Mine is
```bash
> echo $fpath
/Users/dragonly/.oh-my-zsh/plugins/git /Users/dragonly/.oh-my-zsh/functions /Users/dragonly/.oh-my-zsh/completions /Users/dragonly/.oh-my-zsh/cache/completions /usr/local/share/zsh/site-functions /usr/share/zsh/site-functions /usr/share/zsh/5.8.1/functions
```

gaocegege:
I think it should be a bug.

dragonly:
IMO, the owner of `/usr/local/share/zsh/site-functions/` should be root?

gaocegege:
In my case, it's UID1000.

gaocegege:
> Maybe it's better to search for $fpath and find out the proper directory to install the autocomplete scripts?

The solution LGTM!

We need to  find a proper dir.

Triple-Z:
Maybe I can try on this issue.

I think the zsh path can be the following three directories, priorities from high to low:
- `/usr/share/zsh/site-functions`
- `/usr/local/share/zsh/site-functions`
- `$HOME/.oh-my-zsh/completions`

And try to access them each by each until find a proper one.

gaocegege:
Thanks! /assign @Triple-Z

PR #1025, opened 2022-10-13
fix(zsh): ignore inserting zsh-completion if system don't have zsh shell

Triple-Z:
resolves: #552 

- Change envd's zsh-completion script name to `.envd.zsh`
- Try to write the `.envd.zsh` by the following methods:
  1. write `/usr/share/zsh/site-functions/.envd.zsh` directly.
  2. write `/usr/local/share/zsh/site-functions/.envd.zsh` directly.
  3. write `$HOME/.envd.zsh` directly and add loaded script into the `$HOME/.zshrc`

Signed-off-by: Triple-Z <me@triplez.cn>

### DIFF
--- a/pkg/autocomplete/zsh.go
+++ b/pkg/autocomplete/zsh.go
@@ -17,6 +17,7 @@ package autocomplete
 import (
 	"fmt"
 	"os"
+	"os/exec"
 	"os/user"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


pkg/autocomplete/zsh.go
```diff
@@ -52,33 +53,62 @@ _cli_zsh_autocomplete() {
 
 compdef _cli_zsh_autocomplete envd`
 
+var zshConfig = `
+# envd zsh-completion
+[ -f ~/.envd.zsh ] && source ~/.envd.zsh
```
gaocegege, line 58: Please use the path ~/.config/envd/

gaocegege: I am wondering what will happen if we deleted envd?

Triple-Z: If the `.envd.zsh` has been deleted, the condition `[ -f ~/.envd.ssh]` will be false and nothing will happen.

BTW, if the envd has been installed more than once and always use the third method(`~/.envd.zsh` and `~/.zshrc`) on a machine, then the `~/.zshrc` will have multiple same lines of `[ -f ~/.envd.zsh ] && source ~/.envd.zsh`. Fortunately, it will not affect the user, not a big deal.

gaocegege: Gotcha, then I think it should be in `~/.config/envd`

Triple-Z: > Please use the path ~/.config/envd/ 

plus, should we change script's name to `.envd.zsh`? Or just keep the old name `_envd`.

gaocegege: Maybe use envd.zsh.

There is no need to hide it in the file system.

Triple-Z: I changed the code to `~/.config/envd/` directory, think it as resolved.

pkg/autocomplete/zsh.go
```diff
@@ -68,12 +68,12 @@ func InsertZSHCompleteEntry() error {
...
-	filename := ".envd.zsh"
+	filename := "envd.zsh"
 	homeDir := os.Getenv("HOME")
 	dirs := []string{
-		"/usr/share/zsh/site-functions",
-		"/usr/local/share/zsh/site-functions",
-		homeDir,
+		//"/usr/share/zsh/site-functions",
+		//"/usr/local/share/zsh/site-functions",
+		fmt.Sprintf("%s/.config/envd", homeDir),
```
gaocegege, line 76: Let's use `fileutil.DefaultConfigDir`

Triple-Z: fixed.

gaocegege (commented):
/lgtm

Triple-Z:
It seems that the mainline version can't pass the CI too.

Triple-Z:
I have rebased the `main` branch to this branch, the tests passed now, PTAL @gaocegege  @VoVAllen

gaocegege:
LGTM!

PR #1050, opened 2022-10-18
chore(CLI): Move destroy logic from docker to envd engine.

### LINKED ISSUE #1046: chore(CLI): Move destroy logic from docker to envd engine.
## Description

Now the func `Destroy(ctx context.Context, name string) (string, error)` is defined in `pkg/docker/docker.go`

We should move it to pkg/envd/engine.go and implement in pkg/envd/docker.go

---

yvestumushimire:
Fixes #1046

### DIFF
--- a/e2e/cli/up_test.go
+++ b/e2e/cli/up_test.go
@@ -15,14 +15,11 @@
 package cli
 
 import (
-	"context"
-
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
 	"github.com/tensorchord/envd/e2e"
 	"github.com/tensorchord/envd/pkg/app"
-	"github.com/tensorchord/envd/pkg/docker"
 	"github.com/tensorchord/envd/pkg/home"
 )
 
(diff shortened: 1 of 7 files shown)


e2e/cli/up_test.go
```diff
@@ -37,10 +34,6 @@ var _ = Describe("up command", Ordered, func() {
 		envdApp := app.New()
 		err := envdApp.Run(append(baseArgs, "bootstrap"))
 		Expect(err).NotTo(HaveOccurred())
-		cli, err := docker.NewClient(context.TODO())
```
gaocegege, line 40: We need to keep the test logic here, or the test case will fail

yvestumushimire: How can I replicate Destroy method in tests? I saw `dockerEngine` is not exported @gaocegege

gaocegege: You can define Destroy method in env.Engine. Then fake a context

```go
	c := types.Context{Runner: types.RunnerTypeDocker}
	opt := envd.Options{
		Context: c,
	}
	envdEngine, err := envd.New(clicontext.Context, opt)
	if err != nil {
		return errors.Wrap(err, "failed to create envd engine")
	}
```

pkg/app/destroy.go
```diff
@@ -74,13 +75,18 @@ func destroy(clicontext *cli.Context) error {
 		}
 		ctrName = filepath.Base(buildContext)
 	}
+	c := types.Context{Runner: types.RunnerTypeDocker}
```
gaocegege, line 78: Could you please get the context from the home manager? Maybe like this 

https://github.com/tensorchord/envd/blob/main/pkg/app/env_ls.go#L41:#L48

yvestumushimire: Thank you just changed it :)

gaocegege (commented):
/lgtm
/approve

Issue #1027, opened 2022-10-14
feat(envd-server): Implement envd engine interface for envd-server


gaocegege:
## Description

```
type EnvironmentClient interface {
	PauseEnvironment(ctx context.Context, env string) (string, error)
	ResumeEnvironment(ctx context.Context, env string) (string, error)
	ListEnvironment(ctx context.Context) ([]types.EnvdEnvironment, error)
	ListEnvDependency(ctx context.Context, env string) (*types.Dependency, error)
	ListEnvPortBinding(ctx context.Context, env string) ([]types.PortBinding, error)

	CleanEnvdIfExists(ctx context.Context, name string, force bool) error
	// StartEnvd creates the container for the given tag and container name.
	StartEnvd(ctx context.Context, so StartOptions) (*StartResult, error)

	IsRunning(ctx context.Context, name string) (bool, error)
	Exists(ctx context.Context, name string) (bool, error)
	WaitUntilRunning(ctx context.Context, name string, timeout time.Duration) error
}
```

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

VoVAllen:
```
	ListEnvironment(ctx context.Context) ([]types.EnvdEnvironment, error)
	ListEnvDependency(ctx context.Context, env string) (*types.Dependency, error)
	ListEnvPortBinding(ctx context.Context, env string) ([]types.PortBinding, error)
	Exists(ctx context.Context, name string) (bool, error)
	CleanEnvdIfExists(ctx context.Context, name string, force bool) error

gaocegege:
#dibs

gaocegege:
I will try to use https://github.com/containers/image to get the image manifest in envd-server, then add image labels to the pod.

It is to support funcs like `ListEnvDependency(ctx context.Context, env string) (*types.Dependency, error)`

gaocegege:
The code works

```go
package main

import (
	"context"
	"encoding/json"
	"fmt"

	"github.com/containers/image/v5/docker"
	"github.com/containers/image/v5/image"
)

func main() {
	ref, err := docker.ParseReference("//docker.io/gaocegege/test-envd")
	if err != nil {
		panic(err)
	}
	ctx := context.Background()
	src, err := ref.NewImageSource(ctx, nil)
	if err != nil {
		panic(err)
	}
	// b, _, err := src.GetManifest(ctx, nil)
	// if err != nil {
	// 	panic(err)
[...]

PR #1052, opened 2022-10-19
feat: Update env client to support multiple envs

gaocegege:
Ref #1027


gaocegege:
@kemingy PTAL, thanks

kemingy (approved):
LGTM

PR #1085, opened 2022-10-25
feature: support runtime init script

VoVAllen:
support `runtime.init`


VoVAllen:
@gaocegege @kemingy PTAL

pkg/lang/frontend/starlark/runtime/const.go
```diff
@@ -15,9 +15,10 @@
...
-	ruleExpose  = "runtime.expose"
-	ruleDaemon  = "runtime.daemon"
-	ruleEnviron = "runtime.environ"
-	ruleMount   = "runtime.mount"
+	ruleCommand    = "runtime.command"
+	ruleExpose     = "runtime.expose"
+	ruleDaemon     = "runtime.daemon"
+	ruleEnviron    = "runtime.environ"
+	ruleMount      = "runtime.mount"
+	ruleInitScript = "runtime.init_script"
```
VoVAllen, line 23: Is it better to name as `runtime.init`?

kemingy: SGTM

pkg/lang/ir/supervisor.go
```diff
@@ -28,11 +28,14 @@ import (
...
-stdout = "/var/logs/%[1]s_stdout.log"
-stderr = "/var/logs/%[1]s_stderr.log"
+command = """
+%[2]s
+"""
+stdout = "/var/log/horust/%[1]s_stdout.log"
+stderr = "/var/log/horust/%[1]s_stderr.log"
 user = "${USER}"
 working-directory = "${%[3]s}"
+%[4]s
```
kemingy, line 38: I wonder if this is too flexible. Currently, we only need the `start-after`.

Maybe we can switch to a TOML tool to generate it if it becomes more complex in the future.

pkg/lang/ir/supervisor.go
```diff
@@ -60,22 +74,29 @@ func (g Graph) compileEntrypoint(root llb.State) llb.State {
...
+		for i, command := range g.RuntimeInitScript {
+			entrypoint = g.addNewProcess(entrypoint, fmt.Sprintf("init_%d", i), fmt.Sprintf("/bin/bash -c 'set -euo pipefail\n%s'", strings.Join(command, "\n")), nil)
+			deps = append(deps, fmt.Sprintf("init_%d", i))
+		}
+	}
 
 	if g.RuntimeDaemon != nil {
 		for i, command := range g.RuntimeDaemon {
-			entrypoint = g.addNewProcess(entrypoint, fmt.Sprintf("daemon_%d", i), fmt.Sprintf("%s &\n", strings.Join(command, " ")))
+			entrypoint = g.addNewProcess(entrypoint, fmt.Sprintf("daemon_%d", i), fmt.Sprintf("%s &\n", strings.Join(command, " ")), deps)
```
kemingy, line 88: I think we should emphasize their dependency relationship in the document later.

gaocegege: Let's open an issue in envd-docs to keep track.

kemingy: https://github.com/tensorchord/envd-docs/issues/170

gaocegege:
LGTM

gaocegege:
Sorry, I closed it in accident.

pkg/lang/ir/supervisor.go
VoVAllen: We can refactor this once we need more controls here.

examples/dpgen2/build.envd
```diff
@@ -0,0 +1,40 @@
+def install_kubectl_kind():
```
gaocegege, line 1: Should we keep it in envd examples or envdlib?

VoVAllen: It's not general enough now. `sudo docker network connect kind dpgen2` means only project named as dpgen2 will work

gaocegege: Could we make like this:

```
def install_kubectl_kind(environment_name):
```

VoVAllen: Changed to `def install_kubectl_kind(env_name)`. But I still think `kind` is not a general requirement. Thus I don't think we don't need to put it into envdlib

gaocegege (approved):

kemingy (approved):
/lgtm

PR #1104, opened 2022-10-28
fix: cycle import detection in the interpreter

dragonly:


gaocegege:

pkg/lang/frontend/starlark/interpreter.go
```diff
@@ -85,16 +85,11 @@ func (s *generalInterpreter) load(thread *starlark.Thread, module string) (starl
...
-	if e != nil {
-		return e.globals, e.err
-	}
-	if ok {
+	if _, ok := s.cache[module]; ok {
 		return nil, errors.Newf("Detect cycling import during parsing %s", module)
 	}
 
-	s.cache[module] = nil
-
```
kemingy, line 97: Ref to https://github.com/google/starlark-go/blob/22309ac47eac3526f2882553dac3621cb70197c3/starlark/example_test.go#L96-L116

The original implementation should be correct here. It only requires the `s.cache[module] = e` as you added below.

kemingy (changes requested):

pkg/lang/frontend/starlark/interpreter.go
dragonly: Oh I see. We set nil explicitly as a placeholder to indicate that we are starting to exec a new module.

I'll fix that, and add some comments for the purpose of the code block.

dragonly: I have referred to https://github.com/google/starlark-go/blob/22309ac47eac3526f2882553dac3621cb70197c3/starlark/example_test.go#L96-L116.

Also added comments to make the logic easier to follow. PTAL @kemingy

kemingy (commented):
The comments look good to me. I personally prefer the "return early pattern".

cc @gaocegege WDYT

dragonly:
> I personally prefer the "return early pattern".
SGTM. I'll refactor it.

kemingy (approved):
/lgtm

PR #1108, opened 2022-10-28
fix: panic if the user specify entrypoint for non-costom image

### LINKED ISSUE #1099: bug: config.entrypoint is conflicted with runtime.deamon and runtime.init
## Description

Currently the entrypoint has to be `horust`. Rewriting it by `config.entrypoint` will cause problem. We should consider remove this function. @gaocegege @kemingy WDYT

---

kemingy:
* fix #1099


gaocegege (approved):
LGTM :+1:
/lgtm

Issue #896, opened 2022-09-16
feat: Support export buildkit llb graph by envd


VoVAllen:
## Description

Support export LLB graph

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

aseaday:
What format do you expect?

gaocegege:
JSON I think, just like buildctl does. https://github.com/moby/buildkit/tree/master/cmd/buildctl

VoVAllen:
Reference: https://github.com/moby/buildkit/blob/c8d04dbee24d45013b3c36e9b1173bd737ca74b2/cmd/buildctl/debug/dumpllb.go#L83

PR #1124, opened 2022-10-31
feat(debug): Add llb export

gaocegege:
Fix https://github.com/tensorchord/envd/issues/896

```
./bin/envd --debug debug llb -p ./examples/python-basic --dot > a.dot
dot -Tpng a.dot -o a.png
```

a

### DIFF
--- a/pkg/app/debug_llb.go
+++ b/pkg/app/debug_llb.go
@@ -100,7 +100,7 @@ func debugLLB(clicontext *cli.Context) error {
 
 	ops, err := loadLLB(def)
 	if err != nil {
-		errors.Wrap(err, "failed to load LLB")
+		return errors.Wrap(err, "failed to load LLB")
 	}
 
 	if clicontext.Bool("dot") {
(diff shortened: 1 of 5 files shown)


pkg/app/debug_llb.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+	}
+}
+
+func attr(dgst digest.Digest, op pb.Op) (string, string) {
+	switch op := op.Op.(type) {
+	case *pb.Op_Source:
+		return op.Source.Identifier, "ellipse"
+	case *pb.Op_Exec:
+		return strings.Join(op.Exec.Meta.Args, " "), "box"
```
VoVAllen, line 168: Can also include mount information with `op.Exec.Mounts`

VoVAllen (dismissed):
Just realized the cache key is generated at runtime instead of graph build time 😢  Good feature but seems won't solve the cache debugging problem

gaocegege:
> Just realized the cache key is generated at runtime instead of graph build time cry Good feature but seems won't solve the cache debugging problem

The cache is calculated from the build time info, thus you can use the graph to simulate the process.

Ref https://github.com/moby/buildkit/issues/2383

aseaday:
LGTM. To better solve debug problem. Could we have time record on the graph?

gaocegege:
> etter solve debug problem. Could we have time record on the graph?

It is not possible to do that because the graph is static, and the build is actually not executed.

Issue #1111, opened 2022-10-28
feat: support sshd port listening on non-loopback ip 


aseaday:
## Description
The ssh port always listen on 127.0.0.1. But there may be some cases like:
- I want to expose on 0.0.0.0 on eth0 ip for other computer could access

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

gaocegege:
Maybe we can add a new func

```
config.ssh(address="0.0.0.0")
```

WDYT

aseaday:
Good suggesion. But we also need support the  following configurations:
- a xdg config, users may want to all env listen on a specfic host
- cli option, users may get a envdfile from others and want to listen on a spefic host

PR #1128, opened 2022-10-31
feat: expose support listen addr

aseaday:
https://github.com/tensorchord/envd/issues/1111

### DIFF
--- a/pkg/app/create.go
+++ b/pkg/app/create.go
@@ -68,6 +68,11 @@ var CommandCreate = &cli.Command{
 			Value:   sshconfig.GetPrivateKeyOrPanic(),
 			Hidden:  true,
 		},
+		&cli.StringFlag{
+			Name:  "host",
+			Usage: "Assign the host address for environment ssh acesss server listening",
+			Value: envd.Localhost,
+		},
 	},
 	Action: create,
 }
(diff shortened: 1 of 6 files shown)


pkg/app/create.go
```diff
@@ -68,6 +68,11 @@ var CommandCreate = &cli.Command{
 			Value:   sshconfig.GetPrivateKeyOrPanic(),
 			Hidden:  true,
 		},
+		&cli.StringFlag{
+			Name:  "host",
```
gaocegege, line 72: I am wondering if we should make it as a CLI argument, or keep it in build.envd.

WDYT

aseaday: I took a look at the section of `lang.configs`. I think ssh port listening is like a behaviour of enviroment not applications in environment. Secondly, If users want to use this function, mostly because they want to share this environment or use it in remote server. Both seems to be a operation maintain purpose not application modified requirement.

aseaday: That is why I decide to move this function into cli.

gaocegege: SGTM.

/cc @kemingy @VoVAllen

gaocegege:
Maybe we should add a doc page to introduce the feature.

Issue #1133, opened 2022-11-01
bug: failed to support wsl2 backend docker runtime when using GPU


aseaday:
## Description
```
➜  pytorch-profiler git:(feat/add_tensorrt_support) ../../bin/envd up
WARN[2022-11-01T17:32:49+08:00] start the created contrainer envd_buildkitd   container=envd_buildkitd driver=docker-container image="docker.io/moby/buildkit:v0.10.3" socket=envd_buildkitd
INFO[2022-11-01T17:32:52+08:00] manifest is not updated, skip building        tag="pytorch-profiler:dev"
error: GPU is required but nvidia container runtime is not installed, please refer to https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/install-guide.html#docker
```
But we still could use `docker run --gpus all` to start it with gpu support.

## Reproduction
`envd up` in a cuda install envrionment inside a WSL2 ubuntu
## Additional Info
It was introduced by https://github.com/tensorchord/envd/blob/7bf8c00dfd8c34cc5552590b67b1e69eecaf5a51/pkg/envd/docker.go#L287

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritise the issues with the most 👍.

aseaday:
@VoVAllen  I had submit a pr for this issue.

VoVAllen:
Hi,

I tested locally and current envd works well with CUDA+docker inside WSL on my machine. Do you mean that docker in WSL without cuda runtime will cause the problem now?

aseaday:
I had installed Docker Desktop and Nvidia Driver. I didn't install cuda and nvruntime-docker. But according to the stack of docker desktop and nvidia driver. Even though I didn't install nvruntime-docker, I can use cuda device inside docker.

aseaday:
![image](https://user-images.githubusercontent.com/3927355/199380631-cb36d479-30cc-4c05-9daa-711c65744ec3.png)

aseaday:
It is my docker server.

VoVAllen:
And you can use docker with `--gpu` manually but meet problem with envd?

aseaday:
> And you can use docker with `--gpu` manually but meet problem with envd?

Yes, we check if docker support GPU by detect if it is nvruntime. But there is no need to use nvruntime in wsl2 docker desktop. https://www.docker.com/blog/wsl-2-gpu-support-for-docker-desktop-on-nvidia-gpus/

PR #1134, opened 2022-11-01
Bug/add support for wsl2

aseaday:
https://github.com/tensorchord/envd/issues/1133

### DIFF
--- a/pkg/envd/docker.go
+++ b/pkg/envd/docker.go
@@ -580,8 +580,13 @@ func (e dockerEngine) GPUEnabled(ctx context.Context) (bool, error) {
 		return false, errors.Wrap(err, "failed to get docker info")
 	}
 	logrus.WithField("info", info).Debug("docker info")
-	nv := info.Runtimes["nvidia"]
-	return nv.Path != "", nil
+	if strings.HasSuffix(info.KernelVersion, "WSL2") {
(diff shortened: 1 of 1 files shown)


pkg/envd/docker.go
```diff
@@ -581,7 +581,14 @@ func (e dockerEngine) GPUEnabled(ctx context.Context) (bool, error) {
 	}
 	logrus.WithField("info", info).Debug("docker info")
 	nv := info.Runtimes["nvidia"]
-	return nv.Path != "", nil
+	if nv.Path != "" {
+		return true, nil
+	} else if strings.HasSuffix(info.KernelVersion, "WSL2") {
```
gaocegege, line 586: /cc @VoVAllen

pkg/envd/docker.go
```diff
@@ -581,7 +581,14 @@ func (e dockerEngine) GPUEnabled(ctx context.Context) (bool, error) {
...
+		logrus.Warn("We couldn't detect if your runtime support GPU on WSL2, we will continue to run your environment.")
+		return true, nil
```
VoVAllen, line 588: What if user doesn't have GPU with WSL?

aseaday: I raised a issue for docker win team how to detect cuda device.
But for now, users can start environment but without cuda support

VoVAllen (approved):
/lgtm

PR #1135, opened 2022-11-01
feat: add environment name as label

VoVAllen:


pkg/lang/ir/compile.go
```diff
@@ -187,6 +187,7 @@ func (g Graph) Labels() (map[string]string, error) {
 	}
 	labels[types.ImageLabelRepo] = string(repoInfo)
 
+	labels[types.ContainerLabelName] = string(g.EnvironmentName)
```
gaocegege, line 190: The name should be ImageLabelContainerName, I think.

gaocegege: Because the label is on the image.

VoVAllen:
@gaocegege  PTAL

gaocegege (approved):

PR #1146, opened 2022-11-02
ci: Add semantic PR check

terrytangyuan:
Signed-off-by: Yuan Tang <terrytangyuan@gmail.com>

### DIFF
--- a/.github/semantic.yml
+++ b/.github/semantic.yml
@@ -1,0 +1,2 @@
+# Documentation: https://github.com/zeke/semantic-pull-requests
+titleOnly: true


.github/semantic.yml
```diff
@@ -0,0 +1,2 @@
+# Documentation: https://github.com/zeke/semantic-pull-requests
```
tisonkun, line 1: IIRC https://github.com/zeke/semantic-pull-requests GitHub App has been shut down months ago.

You may use https://github.com/Ezard/semantic-prs instead, which keeps the same feature set while having a free service.

tisonkun: When I check new PRs in this repo, I don't see the semantic PR check result.

terrytangyuan: @gaocegege @VoVAllen @kemingy Would you like to install this alternative app? https://github.com/apps/semantic-prs

PR #1148, opened 2022-11-03
feat: Fetch base image metadata

### LINKED ISSUE #1147: bug(build): Environment variables defined in base image are dropped in the new image
## Description

For example:

```
FROM python:3.9-slim

WORKDIR /root
ENV VENV /opt/venv
ENV LANG C.UTF-8
ENV LC_ALL C.UTF-8
ENV PYTHONPATH /root

RUN apt-get update && apt-get install -y build-essential

# Install the AWS cli separately to prevent issues with boto being written over
RUN pip3 install awscli
# Similarly, if you're using GCP be sure to update this command to install gsutil
RUN curl -sSL https://sdk.cloud.google.com/ | bash
ENV PATH="$PATH:/root/google-cloud-sdk/bin"

ENV VENV /opt/venv
# Virtual environment
RUN python3 -m venv ${VENV}
ENV PATH="${VENV}/bin:$PATH"

# Install Python dependencies
RUN pip install flytekit

# This tag is supplied by the build script and will be used to determine the version
# when registering tasks, workflows, and launch plans
ARG tag
ENV FLYTE_INTERNAL_IMAGE $tag
```

We build a image with the dockerfile, then use it in envd:

```
def build():
    base(language="python", image="xx")
```

The env var defined in the image cannot be used in the new image.

## Additional Info

---

gaocegege:
Fix #1147 

### DIFF
--- a/pkg/lang/ir/custom.go
+++ b/pkg/lang/ir/custom.go
@@ -22,7 +22,8 @@ import (
 )
 
 // nolint:unparam
-func (g Graph) compileCustomPython(aptStage llb.State) (llb.State, error) {
+func (g Graph) compileCustomPython(baseStage llb.State) (llb.State, error) {
+	aptStage := g.compileUbuntuAPT(baseStage)
 	pypiMirrorStage := g.compilePyPIIndex(aptStage)
 
 	builtinSystemStage := pypiMirrorStage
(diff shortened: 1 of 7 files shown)


pkg/lang/ir/custom.go
```diff
@@ -43,7 +44,7 @@ func (g Graph) compileCustomPyPIPackages(root llb.State) llb.State {
 	// Compose the package install command.
 	var sb strings.Builder
 	// Always use the conda's pip.
-	sb.WriteString("pip install")
+	sb.WriteString(`pip install`)
```
kemingy, line 47: Does this one require the absolute path?

gaocegege: We should not use an absolute path here since users may use conda/venv to install a new python environment.

kemingy (approved):
/lgtm

PR #1173, opened 2022-11-10
fix: Move user to image config

gaocegege:


kemingy (approved):
LGTM. But I think the envd-server error is not caused by the user.

gaocegege:
> LGTM. But I think the envd-server error is not caused by the user.

Yep, I verified that it is not from the user.

Issue #706, opened 2022-08-02
feat(CLI): Add a demo command


gaocegege:
## Description

`envd demo` creates a build.envd demo and up it.

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

kemingy:
Can we support `envd up -f <URL>` to build and run the remote envd file?

gaocegege:
SGTM but the build context is still a problem. We cannot get the remote build context from the URL.

VoVAllen:
Probably we can make this part of `envd init` command, such as `envd init --demo`

gaocegege:
In my opinion, the command `envd demo` should create the build.envd in tmp/xx/ then run `envd up` based on this. Then the users can get a shell from it directly.

aseaday:
How about we provide a interfact shell like:
```bash
$ envd demo
> Which one do you want to use: Python/R/Julia
The following packages will be installed for you:
- numpy
- 
> What packages do you want install extra
```

gaocegege:
Looks promising to me. :tada:

gaocegege:
It also can be used in `envd init`

aseaday:
> It also can be used in `envd init`

yes, we could had a feature to support interactive set simple envd build manifest first.

kemingy:
interactive shell:

* https://github.com/charmbracelet/gum

AlexXi19:
This would be helpful, an interactive shell to choose a simple initial configuration.

gaocegege:
I marked the issue priority/medium

AlexXi19:
Interested in this issue, could use this library: https://github.com/charmbracelet/bubbletea

gaocegege:
The lib LGTM. Please pick it up if you are interested in it!

PR #1210, opened 2022-11-18
feat: Interactive cli to generate `build.envd`

AlexXi19:
https://github.com/tensorchord/envd/issues/706

## Interactive CLI to generate `build.envd` (Proof of Concept)

- Asks user questions and generates `build.envd` based on the user's answers

- Used [bubbletea](https://github.com/charmbracelet/bubbletea) to create the TUI and [bubbles](https://github.com/charmbracelet/bubbles) to style the TUI. 

- Supports single selection and multiple selection. 

## Interface for the developers 
Create a data structure to define a new question
```golang
func LanguageChoice() input {
	return input{
		prompt:    "Choose a programming language",
		inputType: SINGLE_SELECT,
		label:     LabelLanguage,
		options: []string{
			"python",
			"r",
			"julia",
		},
	}
}
```  
Add question to `demoCommand`, further define question flow using conditional statements. The user's selection will be added to `selectionMap` using your defined label. 
```golang 
	startQuestion(LanguageChoice())
	if selectionMap[LabelLanguage][0] == "python" {
		startQuestion(PythonPackageChoice())
	} else if selectionMap[LabelLanguage][0] == "r" {
		startQuestion(RPackageChoice())
	}
```

Use the user's selection from `selectionMap` and add your codegen logic to `generateFile`

## Demo 
ezgif-2-ad21c83847

## Testing 
Can add testing later after finalizing the choices.


gaocegege:
Cool!

pkg/app/app.go
```diff
@@ -97,6 +97,7 @@ func New() EnvdApp {
 		CommandDebug,
 		CommandVersion,
 		CommandTop,
+		CommandDemo,
```
gaocegege, line 100: How about reusing `envd init`?

pkg/app/demo.go
```diff
@@ -0,0 +1,301 @@
...
+
+const (
+	LabelLanguage      string = "Language"
+	LabelCuda          string = "Cuda"
+	LabelCudaChoice    string = "Cuda Choice"
+	LabelPythonPackage string = "Python Package"
+	LabelRPackage      string = "R Package"
+)
+
+type model struct {
```
gaocegege, line 86: Maybe we can create a new package `app/interactive` for the implementation, to increase maintainability

gaocegege:
/assign @kemingy @VoVAllen 

pkg/app/demo.go
```diff
@@ -0,0 +1,301 @@
...
+	_, err := p.Run()
+	if err != nil {
+		fmt.Printf("There was an error generating build.envd: %v", err)
+		os.Exit(1)
+	}
+}
+
+var selectionMap = make(map[string][]string)
+var itemStyle = lipgloss.NewStyle().Foreground(lipgloss.Color("#999999"))
+var selectedItemStyle = lipgloss.NewStyle().Foreground(lipgloss.Color("#4dff4d"))
```
kemingy, line 70: Need to consider the light theme.

pkg/app/demo.go
```diff
@@ -0,0 +1,301 @@
...
+
+func CudaVersionChoice() input {
+	return input{
+		prompt:    "Choose a cuda version",
+		inputType: SINGLE_SELECT,
+		label:     LabelCuda,
+		options: []string{
+			"11.0",
+			"10.2",
+			"10.1",
```
kemingy, line 157: Please update the version according to the https://github.com/tensorchord/envd/blob/main/base-images/remote-cache/build.envd

pkg/app/app.go
AlexXi19: I am fine with what you decide, I created a new command just in case.

pkg/app/demo.go
AlexXi19: Yep, I didn't do it because im still getting a feel of the code style and convention. I can do this.

pkg/app/demo.go
AlexXi19: Yes, lipgloss has options for this, i will add this. 

https://github.com/charmbracelet/lipgloss#adaptive-colors

pkg/app/demo.go
AlexXi19: Ok, will do!

AlexXi19:
Compiling a list of decisions to be finalized: 

1. Which cli command do we want to use for this? 
2. I only put in some sample questions for now, would like to get a final confirmation for the prompt questions and options that we want to support (and the exact verbiage). 
3. Final UI/UX changes: cursor, selection color, navigation & selection button commands. 
4. Other edge cases such as when `build.envd` already exist, throw error or overwrite. 
5. Any other things we want to add.

cc: @gaocegege @kemingy

aseaday:
> Compiling a list of decisions to be finalized:
> 
> 1. Which cli command do we want to use for this?
> 2. I only put in some sample questions for now, would like to get a final confirmation for the prompt questions and options that we want to support (and the exact verbiage).
> 3. Final UI/UX changes: cursor, selection color, navigation & selection button commands.
> 4. Other edge cases such as when `build.envd` already exist, throw error or overwrite.
> 5. Any other things we want to add.
> 
> cc: @gaocegege @kemingy

1.`envd init` is good to me.
3. throw error, because overwriting sometimes will cause data lost.

kemingy:
> Compiling a list of decisions to be finalized:
> 
>     1. Which cli command do we want to use for this?
`envd init` looks good to me. If the work path already has `requirements.txt` or `env.yaml`, I guess we can skip the python packages step. 
>     2. I only put in some sample questions for now, would like to get a final confirmation for the prompt questions and options that we want to support (and the exact verbiage).
Maybe can add the jupyter or tensorboard, etc.
>     3. Final UI/UX changes: cursor, selection color, navigation & selection button commands.
I don't have opinions about this. As long as it's visible under the dark/light theme.
>     4. Other edge cases such as when `build.envd` already exist, throw error or overwrite.
To make it easy, just throw errors.
>     5. Any other things we want to add.
I wonder if we should add some commented code like the following block. This can remind users we have such features.

```py
# envdlib = include("https://github.com/tensorchord/envdlib")

def build():
    # runtime.expose(envd_port=8000, host_port=8000, service="demo")
    # envdlib.tensorboard(host_port=8888)
```

aseaday (approved):

kemingy:
> The default color schemes in applications must be restricted to 12 colors: red, green, yellow, blue, magenta, cyan, and the bright versions of each of these.

* https://rust-cli-recommendations.sunshowers.io/colors.html

AlexXi19:
@kemingy 

> envd init looks good to me. If the work path already has requirements.txt or env.yaml, I guess we can skip the python packages step.

What do you mean by this? Do you mean removing package question if there's `requirements.txt` or `env.yaml` or generating `build.env` by reading in `requirements.txt`? If the former, does `envd` read in `requirements.txt` while building the image?

kemingy:
> @kemingy
> 
> > envd init looks good to me. If the work path already has requirements.txt or env.yaml, I guess we can skip the python packages step.
> 
> What do you mean by this? Do you mean removing package question if there's `requirements.txt` or `env.yaml` or generating `build.env` by reading in `requirements.txt`? If the former, does `envd` read in `requirements.txt` while building the image?

Skip the package question. No need to read in, not sure about your concern.

> Another question, should I remove all the current code in `init.go` and put the code for the interactive cli there instead?

No need. They are not in conflict.

AlexXi19:
New Revision: 
- Refactor interactive cli logic to new file 
- Accommodate for terminal color
- Merge new interactive CLI logic with existing `init` logic that reads relevant files from the directory

gaocegege:
Please fix the e2e test.

The code LGTM

/assign @kemingy @VoVAllen 

Thanks for the contribution!

PR #1228, opened 2022-11-25
feat: support serving

kemingy:
Breaking changes:
- [x] `base(image, dev)`
- [x] `install.conda()`
- [x] `install.python()`
- [ ] `install.r_lang()` not implemented yet
- [ ] `install.julia()` not implemented yet
- [ ] `config.jupyter()` move to envdlib (TBD)
- [ ] `config.rstudio_server()` move to envdlib (TBD)


kemingy:
* conda is required if Python needs to be installed
  * I checked pyenv but it requires a dev env to compile from source
* CUDA only works if the `image == defaultImage`
* it seems serving env also needs some system packages like `wget` to install conda, although this can be done by LLB, WDYT
* I plan to move `jupyter` and `rstudio` to envdlib, WDYT

cc @gaocegege @VoVAllen It's not 100% finished, but I think you can take a look

kemingy:
To test this PR, you can try:

```py
def build():
    dev_tools()
    install.conda()
    install.python()
    install.python_packages(name=["ipython"])
    install.apt_packages(name=["ripgrep"])
```

gaocegege:
I think we can keep jupyter since it is used frequently. Maybe we could have a new package `editor`?

gaocegege:
We should schedule v0.3.0 for the PR. Since it is a breaking change to user interface. WDYT

pkg/lang/ir/compile.go
```diff
@@ -114,7 +112,7 @@ func CompileEntrypoint(buildContextDir string) ([]string, error) {
 }
 
 func CompileEnviron() []string {
-	if DefaultGraph.Image != nil {
+	if !DefaultGraph.DevTools {
```
gaocegege, line 115: Is there a better name for `dev_tools()`?

kemingy:
> We should schedule v0.3.0 for the PR. Since it is a breaking change to user interface. WDYT

Yep.



> I think we can keep jupyter since it is used frequently. Maybe we could have a new package `editor`?

* people should prefer jupyter-lab, if you visit https://jupyter.org/, it shows "JupyterLab: A Next-Generation Notebook Interface"
* I think they are still daemon processes, we don't need to keep too many specific codes

pkg/lang/ir/compile.go
kemingy: Any suggestions?

VoVAllen:
I think we can move jupyter to envdlib. If it's not convenient enough for user, we can embed envdlib and include it automatically. Also what's our plan if install.python is called without install.conda?

kemingy:
> Also what's our plan if install.python is called without install.conda?

For now, it will still install conda. I didn't find a "portable Python". `pyenv` will download the source code and compile it. Will check miniconda source code later.

gaocegege:
> I think we can move jupyter to envdlib. If it's not convenient enough for user, we can embed envdlib and include it automatically.

If we embed the envdlib, there is no difference between envdlib and `install`

pkg/lang/ir/compile.go
gaocegege: - enable_development()
- dev_env()

BTW, should we enable dev tools by default? Or disable it by default? Here are two different designs.

```
def build():
    dev_tools()
```

```
def serve():
    disable_dev()

Comment (kemingy) [reply pkg/lang/ir/compile.go:115]: Usually, a function is used to enable some features.

enable_development() is too long to type. dev_env() looks okay, but it actually doesn't add produce an env. Instead, it adds some tools and configurations.

**Comment by kemingy** [comment]
* Python without conda can be done by
  * copy micromamba
  * create a Python env
  * rm micromamba
* Shall we install "curl ca-certificates bzip2" for serving? I feel copy the `/etc/ssl/certs/ca-certificates.crt` is not elegant.
* Not sure which `user` should we put in `ImageConfig.User`
* Need to remove the hard-code sshd label

**Comment by gaocegege** [comment]
> Shall we install "curl ca-certificates bzip2" for serving? I feel copy the /etc/ssl/certs/ca-certificates.crt is not elegant.

I think it should be handled by users. And we do not need to copy certs to it.

**Comment by gaocegege** [comment]
> Not sure which user should we put in ImageConfig.User

Could we not care about the user? It only be added with `dev_tools`

**Comment by kemingy** [comment]
> > Shall we install "curl ca-certificates bzip2" for serving? I feel copy the /etc/ssl/certs/ca-certificates.crt is not elegant.
> 
> I think it should be handled by users. And we do not need to copy certs to it.

No, we have to. Without the certs, mamba cannot download the required libraries to create a Python env. And you can get the certs if you install `ca-certificates`.

Refer to: https://github.com/mamba-org/micromamba-docker/blob/903975a7cc10f20d8cab684e5a4f9df6485279b4/Dockerfile#L27



> > Not sure which user should we put in ImageConfig.User
> 
> Could we not care about the user? It only be added with `dev_tools`

Now it hard codes as `envd`. I will try to find a way to access the DefaultGraph there.

**Comment by kemingy** [comment]
> > Shall we install "curl ca-certificates bzip2" for serving? I feel copy the /etc/ssl/certs/ca-certificates.crt is not elegant.
> 
> I think it should be handled by users. And we do not need to copy certs to it.

BTW, without these libraries, we will need to use `curlimages/curl:latest` or `buildpack-deps:stable-curl` to download and unpack miniconda.

**Comment by kemingy** [comment]
> Breaking changes:
> 
>     * [x]  `base()`
> 
>     * [x]  `dev_tools()`
> 
>     * [x]  `install.conda()`
> 
>     * [x]  `install.python()`
> 
>     * [ ]  `install.r_lang()` not implemented yet
> 
>     * [ ]  `install.julia()` not implemented yet
> 
>     * [ ]  `config.jupyter()` move to envdlib (TBD)
> 
>     * [ ]  `config.rstudio_server()` move to envdlib (TBD)

Check the progress on the top of this PR. I prefer to implement the rest in other PRs.

It's ready to review. cc @gaocegege @VoVAllen 

A bug needs your help: the envd user is not added to the sudo group. I don't know why.

You can try it with:

```py
def build():
    dev_tools()
    install.python()
    install.python_packages(name=["via"])
    install.conda()
    install.conda_packages(name=["exa"])


def serving():
    install.conda(use_mamba=True)
    install.python(version="3.9")
    install.python_packages(name=["via"])
    install.conda_packages(name=["exa"])
```

The serving image needs to run with `docker run` because it doesn't have envd-sshd.

kemingy:
### Update

Deleted `dev_tools()`, adding `dev` as a parameter to `base`.

```py
def build():
    # dev
    base(dev=True)
    ...

    # serving
    ...
```

Changed to use `curlimages/curl` to download miniconda. This one is tiny.

Fixed user group bug and vscode merge bug.

PTAL @gaocegege @VoVAllen

gaocegege:
May I ask why e2e test fails?

kemingy:
> May I ask why e2e test fails?

I haven't changed the envd file under the test folder. So it should always fail.

If you're okay with the code change, I can update all the test files.

gaocegege:
https://github.com/tensorchord/envd/pull/1228#issuecomment-1328419486

Could you please update the latest demo here?

gaocegege:
> If you're okay with the code change, I can update all the test files.

The code change LGTM.

PR #1272, opened 2022-12-06
chore: Keep v1 and v0 in envd

gaocegege:


aseaday:
A good signal to step forward to a more promising future of API. But could we have a guide for developers a whole view of v2?

gaocegege:
> A good signal to step forward to a more promising future of API. But could we have a guide for developers a whole view of v2?

I think we should have such a guide. Maybe after #1228 is merged

/cc @kemingy

pkg/app/create.go
```diff
@@ -27,7 +27,7 @@ import (
 
 	"github.com/tensorchord/envd/pkg/envd"
 	"github.com/tensorchord/envd/pkg/home"
-	"github.com/tensorchord/envd/pkg/lang/ir"
+	v1 "github.com/tensorchord/envd/pkg/lang/ir/v0"
```
kemingy, line 30: Why alias `v1 = v0`?

pkg/lang/frontend/starlark/v0/starlark_suite_test.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package starlark
+package v1
```
kemingy, line 15: v0

pkg/lang/frontend/starlark/v1/testdata/test.envd
```diff
@@ -0,0 +1,8 @@
+def build():
+    base(os="ubuntu20.04", language="python3")
```
kemingy, line 2: This is a v0 test file.

I think we don't really run this test.

gaocegege: https://github.com/tensorchord/envd/blob/main/pkg/lang/frontend/starlark/testdata/test.envd 

The code is also v0 in upstream/main.

But it does not matter because the test just gets the hash

```go
var _ = Describe("Starlark", func() {
	It("should be able to compile envd", func() {
		filename := "testdata/test.envd"
		hash, err := GetEnvdProgramHash(filename)
		Expect(err).NotTo(HaveOccurred())
		Expect(hash).To(Equal("cff1c81818116d42"))
	})
})
```

kemingy: I see. But I think we should change it to a v1 file. In case we don't support v0 in the future, then the interpreting will fail.

gaocegege: I already updated it.

gaocegege:
Comments are addressed, please take a look.
Thanks for your time :smile:

kemingy:
CI report failed because it should `needs: v1...`. @gaocegege

kemingy (approved):

kemingy:
Will fix the CI test coverage report in another PR.

VoVAllen:
tql

PR #1273, opened 2022-12-07
Remove incorrect `sudo` command when installing system package

oubotong:
In the function `compileSystemPackages()`, the `sudo` before `apt-get install` prevent the base environment configuration from taking effect (such as `DEBIAN_FRONTEND=noninteractive` setting). 

This would result in the container being blocked by some package installation at booting time. For example, when installing `tzdata`:
d6a70383b0c3b244e62553aba12c51e

Since we are already root during the booting time, there is also no need to add `sudo`  there. 

To reproduce, try the following `build.env`:
```
def build():
    # config.pip_index(url = "https://pypi.tuna.tsinghua.edu.cn/simple")
    install.apt_packages(name = [
        "tzdata",
    ])
    shell("zsh")
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -30,7 +30,7 @@ require (
 	github.com/sirupsen/logrus v1.9.0
 	github.com/spf13/viper v1.14.0
 	github.com/stretchr/testify v1.8.1
-	github.com/tensorchord/envd-server v0.0.12
+	github.com/tensorchord/envd-server v0.0.13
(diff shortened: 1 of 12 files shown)


kemingy:
Please rebase the tensorchord/envd:main.

oubotong:
> Please rebase the tensorchord/envd:main.

done

kemingy (approved):
LGTM. cc @gaocegege

gaocegege:
@allcontributors Please add @oubotong for the code contribution!

PR #1280, opened 2022-12-07
WIP add docs about support version

aseaday:
I run a bunch of tests  of cuda and os test on envd. here is part of results. Because it is a little confusing to know cuda support.
Please gimme some suggestions about support document. such as:

- for users install pytorch with cuda, do we need a cuda image?

When I finish the serving PR and newsletter work, I will add a CI to it.

### DIFF
--- a/docs/support.md
+++ b/docs/support.md
@@ -1,0 +1,52 @@
+# Support
+
+The following features are supported by envd team.
+
+> envd v0.2.5
+
+## OS support
+
+- ubuntu 20.04
+
+## CUDA
+
+Within the following support cuda version, you could use:
+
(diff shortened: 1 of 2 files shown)


docs/support.md
```diff
@@ -0,0 +1,52 @@
+# Support
+
+The following features are supported by envd team.
+
+> envd v0.2.5
+
+## OS support
+
+- ubuntu 20.04
```
kemingy, line 9: We are able to support any Debian-based image now (with `# syntax=v1`)

aseaday: marked and I will have a try next week.

PR #1282, opened 2022-12-08
chore: Refine CLI interfaces

gaocegege:
Ref https://github.com/tensorchord/envd/discussions/1279


pkg/app/run.go
```diff
@@ -16,140 +16,200 @@ package app
...
-		}
-		// Get the environment name if `name` is not specified.
-		if name == "" {
-			name = filepath.Base(path)
-		}
+	if c.Runner == types.RunnerTypeDocker {
+		return errors.Newf("docker runner is not supported for this command, please use `envd up`")
 	}
+	telemetry.GetReporter().Telemetry(
+		"run", telemetry.AddField("runner", c.Runner))
```
kemingy, line 105: Shall we collect how much time the envd-server takes to create the environment? If so, we should add the duration from the beginning of this function to the detach part.

gaocegege: It is not accurate now. For example, I always ctrl c when running run. Then I watch the k8s cluster and ssh when it is ready.

kemingy (approved):

PR #1284, opened 2022-12-08
doc: v1

kemingy:

cc @VoVAllen 

It's still ugly. The class method has exactly the same indent as a normal function or a class object. Not sure if we can fix it.


VoVAllen:
I think it's set at https://github.com/tensorchord/envd-docs/blob/main/hack/api/render.py#L37-L43

VoVAllen:
You can change it if needed

kemingy:
> I think it's set at https://github.com/tensorchord/envd-docs/blob/main/hack/api/render.py#L37-L43

Yes, but the header level is not sufficient. I wonder if this can generate the indents like the Python sphinx doc: https://0b01001001.github.io/spectree/plugins.html

We can merge this one and find a better way letter for the envd-doc repo.

VoVAllen:
I think we don't need to make them as sub class. Does sub module work?

kemingy:
This change should work with https://github.com/tensorchord/envd-docs/pull/206.

I hope we can add the module path to the function title, but sadly there is a bug in the https://github.com/NiklasRosenstein/pydoc-markdown/issues/276

gaocegege:
Could you please give a preview link here?

gaocegege:
About the markdown or html generated by the v1 API

kemingy:
image

kemingy:
Screenshot updated. I guess we can also patch the pydoc-markdown if necessary.

gaocegege:
I am wondering where to tell users that the default version is v0, and they can use `# syntax=v1` to move to v1.

kemingy:
> I am wondering where to tell users that the default version is v0, and they can use `# syntax=v1` to move to v1.

I added to every header of the v1 module doc. See the warning in the screenshot.

Juni66623:
Thu

Issue #410, opened 2022-06-17
feat(lang): Support UID/GID as a ARG in build.envd (just like vscode dev container does)


gaocegege:
@gaocegege if envd can remapping `UID` and `GID` at runtime that will be great. I'm also agree with @Zheaoli that if UID/GID remapping is doable there should be an API or config option allowing user to modify it.

I've used VSCode Dev Container with some projects, and I found that they also need to fix this UID/GID issue. https://github.com/microsoft/vscode-dev-containers/blob/main/containers/python-3/.devcontainer/library-scripts/common-debian.sh#L182

As far as I understand it, they fix it in build time but not build from scratch. The base Dockerfile has several build args. The user firstly downloads the base image and re-builds it with new args e.g., NEW_UID, NEW_GID.

Here is the build log I captured from VSCode console:

```bash
[11184 ms] Start: Check Docker is running
[11184 ms] Start: Run in Host: docker version --format {{.Server.APIVersion}}
[11405 ms] Server API version: 1.41
[11408 ms] Start: Run in Host: docker volume ls -q
[11581 ms] Start: Run in Host: docker inspect --type container 016e7f0fda022020f0e1403ffb8c970ed625c157df31e85f46c4c69a950561ac
[11649 ms] Start: Run in Host: docker rm -f 016e7f0fda022020f0e1403ffb8c970ed625c157df31e85f46c4c69a950561ac
[12280 ms] Start: Run in Host: docker ps -q -a --filter label=vsch.local.folder=/home/aaron/tmp/devcontainer-test --filter label=vsch.quality=stable
[12346 ms] Start: Run in Host: docker ps -q -a --filter label=devcontainer.local_folder=/home/aaron/tmp/devcontainer-test
[...]

PR #1286, opened 2022-12-08
Add config.owner to v1 syntax for setting UID and GID

aseaday:
This is a extra pr for #410.
There three ways to set UID/GID during building:

1. read args from cli
2. config.owner in build.envd
3. get current user's uid/gid
and their priorities.

### DIFF
--- a/pkg/lang/frontend/starlark/v1/config/config.go
+++ b/pkg/lang/frontend/starlark/v1/config/config.go
@@ -45,6 +45,7 @@ var Module = &starlarkstruct.Module{
 		"rstudio_server": starlark.NewBuiltin(ruleRStudioServer, ruleFuncRStudioServer),
 		"entrypoint":     starlark.NewBuiltin(ruleEntrypoint, ruleFuncEntrypoint),
 		"repo":           starlark.NewBuiltin(ruleRepo, ruleFuncRepo),
+		"owner":          starlark.NewBuiltin(ruleOwner, ruleFuncOwner),
 	},
 }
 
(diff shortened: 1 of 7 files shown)


gaocegege (approved):
/cc @kemingy 

LGTM

pkg/lang/frontend/starlark/v1/config/config.go
```diff
@@ -225,3 +226,24 @@ func ruleFuncRepo(thread *starlark.Thread, _ *starlark.Builtin,
...
+
+	if err := starlark.UnpackArgs(ruleOwner, args, kwargs, "uid", &envdUid, "gid?", &envdGid); err != nil {
+		return nil, err
+	}
+	uid, _ := envdUid.Int64()
+	gid, _ := envdGid.Int64()
+	if uid < 0 || uid > 65535 || gid < 0 || gid > 65535 {
+		err := errors.New("get a wrong uid or gid")
+		return nil, err
+	}
```
kemingy, line 245: If `gid` is optional, seems it will fail because `gid=-1` by default.

aseaday: I think I should modify the optional properity.

kemingy (approved):
LGTM.

Shall we support this feature in v0? @gaocegege

gaocegege:
Let's merge it and see. I think there is no need to support for v1 now. WDYT

aseaday:
I don't think we need to support it in v1. I think all v0 update from today should be only bug fix or critical issues.

PR #1295, opened 2022-12-12
fix: add PATH when switch to envd user

kemingy:

* related to https://github.com/tensorchord/envd/pull/1273

We should re-add the env


pkg/lang/ir/v1/user.go
```diff
@@ -32,7 +34,12 @@ func (g *generalGraph) compileUserOwn(root llb.State) llb.State {
 		root = root.Run(llb.Shlex(fmt.Sprintf("chown -R envd:envd %s", dir)),
 			llb.WithCustomNamef("[internal] configure user permissions for %s", dir)).Root()
 	}
-	return root.User("envd")
+	user := root.User("envd")
+	// re-add the env since it's a different user
+	for _, env := range types.BaseEnvironment {
+		user = user.AddEnv(env.Name, env.Value)
```
gaocegege, line 40: Will this break cache?

PR #1296, opened 2022-12-13
fix: make starship prompt pure text

kemingy:

Refer: https://starship.rs/presets/plain-text.html


gaocegege (approved):
/lgtm

PR #1300, opened 2022-12-13
doc: add more installation methods

kemingy:

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,10 +111,17 @@ For example, the PyPI cache is shared across builds and thus the package will be
 
 ### Install and bootstrap `envd`
 
-`envd` can be installed with `pip` (only support Python3). After the installation, please run `envd bootstrap` to bootstrap.
+`envd` can be installed with `pip`, or you can download the binary [release](https://github.com/tensorchord/envd/releases) directly. After the installation, please run `envd bootstrap` to bootstrap.
 
 ```bash
-pip3 install --pre --upgrade envd
+# install with local pip
+pip install --pre --upgrade envd
+# or install with pipx
+pipx install envd --pip-args="--pre"
+# or download the latest release artifact with GitHub CLI
+bash -c "gh release download -R tensorchord/envd -p $(echo envd_\*_$(uname)_$(uname -p))"
+
+# bootstrap
 envd bootstrap
 ```
 


README.md
```diff
@@ -111,10 +111,17 @@ For example, the PyPI cache is shared across builds and thus the package will be
...
+`envd` can be installed with `pip`, or you can download the binary [release](https://github.com/tensorchord/envd/releases) directly. After the installation, please run `envd bootstrap` to bootstrap.
 
 ```bash
-pip3 install --pre --upgrade envd
+# install with local pip
+pip install --pre --upgrade envd
+# or install with pipx
+pipx install envd --pip-args="--pre"
+# or download the latest release artifact with GitHub CLI
+bash -c "gh release download -R tensorchord/envd -p $(echo envd_\*_$(uname)_$(uname -p))"
```
gaocegege, line 122: What will happen if there is no gh installed?

kemingy: Why will you run this command if you don't have gh installed?

gaocegege: because I do not know what the command means from a ds's perspective.

kemingy: See the comments, `with GitHub CLI`.

I don't think you will run `pipx` if you don't have `pipx`. The same to `gh`.

frostming:
It surprises me that there is no 'installation' section in the docs, or is there one that I missed?

IMHO we should list recommended installation methods together with alternatives in the docs, and keep the installation part in README simple.

kemingy:
> It surprises me that there is no 'installation' section in the docs, or is there one that I missed?
> 
> IMHO we should list recommended installation methods together with alternatives in the docs, and keep the installation part in README simple.

The current [installation section](https://github.com/tensorchord/envd#install-and-bootstrap-envd) doesn't have enough instructions for beginners. Users may not know "how to add it to the PATH" or "why I'm using the old one even after `pip install -U --pre envd`".

I agree we should improve the installation section.

frostming:
@kemingy Yeah, we can add an installation section to [Building Your First Environment page](https://envd.tensorchord.ai/guide/build-envd.html), or in its own page. We can then make it more verbose and clear there, and leave a link referring to the docs in README.

gaocegege:
BTW, should we remove --pre from the script? Since we will follow the semver to publish new releases.

kemingy:
> BTW, should we remove --pre from the script? Since we will follow the semver to publish new releases.

Agree. cc @frostming you can delete `--pre` when you're trying to work on https://github.com/tensorchord/envd-docs/issues/204

PR #1312, opened 2022-12-16
chore: Use pip by default in readme

gaocegege:


gaocegege:
/cc @frostming 

The installation speed is too slow with install.sh. Let's use pip by default.

frostming (approved):

PR #1333, opened 2022-12-22
feat(login): Refactor logic with the new key API

gaocegege:
Since we added a new API group about ssh key, this PR is to refactor the envd to support it.


kemingy (approved):
/lgtm

PR #1337, opened 2022-12-23
refact: change Shlex(fmt.Sprintf to Shlexf, use ` instead of escape \"

kemingy:
Signed-off-by: Keming <kemingy94@gmail.com>


gaocegege (commented):
/lgtm

PR #1352, opened 2022-12-29
File sync for envd-server

AlexXi19:
Draft pull request for https://github.com/tensorchord/envd-server/issues/102


pkg/syncthing/client.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+const (
+	GET  = "GET"
+	POST = "POST"
+	PUT  = "PUT"
+)
+
+// Makes API calls to the syncthing instance
+func (s *Syncthing) ApiCall(method string, url string, params map[string]string, body []byte) ([]byte, error) {
```
VoVAllen, line 25: Is this code borrowed from other repo? Please keep license head if so.

VoVAllen: Also the client style is not consistent with our current HTTP client

VoVAllen (commented):
The client part seems need more discussions

pkg/syncthing/client.go
AlexXi19: Ok, yeah this is borrowed for now, I was going to ask you about the http client after you wake up.

PR #1373, opened 2023-01-03
fix(CLI): Remove image list in destroy

### LINKED ISSUE #1372: bug(CLI): destroy command is broken
## Description

```
$ ./bin/envd d --name mnist                      
error: json: cannot unmarshal array into Go value of type string
```

## Additional Info

---

gaocegege:
Fix #1372 

Ref https://github.com/tensorchord/envd/commit/0a975de0103d84841252b54b549cddaa6d82d8a6#diff-e2ddd9833874ac79a5e345e2cdb9293b6fa55ab86269dc103de3b27c9288bb77R520

In this func a image list is run. Then it will return an error if there is any broken image.
I run envd destroy --name mnist, it returns an error because python-basic is broken.


kemingy (dismissed):
/lgtm

PR #1375, opened 2023-01-03
fix: change to dynamic PATH

### LINKED ISSUE #1346: refactor: `DefaultPathEnvUnix` should be dynamic
## Description

Currently, the `DefaultPathEnvUnix` is hard-coded with conda path and Julia path. Since we support different languages, this should be a dynamic one. It should not include the Julia path for Python environments.

---

kemingy:
* fix #1346 

## Usage

* runtime env
* build time path
  * init
  * renew when switch the user

## Design

* maintain it in the Graph (RuntimeEnvPaths)
* init as the system PATH
* update if needed

---

*Note*: runtime env PATH leaks to build time

### DIFF
--- a/pkg/lang/ir/v0/system.go
+++ b/pkg/lang/ir/v0/system.go
@@ -54,7 +54,6 @@ func (g generalGraph) compileRun(root llb.State) llb.State {
 	}
 
 	workingDir := g.getWorkingDir()
-	stage := root
 	for _, execGroup := range g.Exec {
 		var sb strings.Builder
 		sb.WriteString("set -euo pipefail\n")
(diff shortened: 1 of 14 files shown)


pkg/lang/ir/v0/system.go
```diff
@@ -54,7 +54,7 @@ func (g generalGraph) compileRun(root llb.State) llb.State {
 	}
 
 	workingDir := g.getWorkingDir()
-	stage := root.AddEnv("PATH", types.DefaultPathEnvUnix)
+	stage := root
```
gaocegege, line 57: There is no need to keep stage.

pkg/lang/ir/v1/julia.go
```diff
@@ -21,9 +21,11 @@ import (
 	"github.com/cockroachdb/errors"
 	"github.com/moby/buildkit/client/llb"
 	"github.com/sirupsen/logrus"
+	"github.com/tensorchord/envd/pkg/types"
 )
 
 func (g generalGraph) installJulia(root llb.State) (llb.State, error) {
```
gaocegege, line 28:
```suggestion
func (g *generalGraph) installJulia(root llb.State) (llb.State, error) {
```

pkg/lang/ir/v0/system.go
kemingy: Fixed

gaocegege (approved):
LGTM :+1:
/lgtm

Issue #576, opened 2022-07-08
feat: containerd/nerdctl provider support


junnplus:
## Description

[nerdctl](https://github.com/containerd/nerdctl/) is a Docker-compatible CLI for [containerd](https://containerd.io/) and it has many new features like rootless mode, compose integration, work with buildkit, and [so on](https://github.com/containerd/nerdctl/#features-present-in-nerdctl-but-not-present-in-docker).

PS. I am one of the main maintainers of containerd/nerdctl.

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

aseaday:
LGTM,  I will follow this issue.

aseaday:
I look throught the code of nerdctl, and there may be a little problem:
![image](https://user-images.githubusercontent.com/3927355/177960554-d0ab0d89-df70-45a9-b706-fa95af31aded.png)
As you can see from the illustrations,  envd use the golang's docker client package to talk with the dockerd.  There seems to be no such a thing in nerdctl.  A way to solve this I think is that you could provide a client sdk which work the same way as cli binary. Then nerdctl users could build image and container by `envd` and see the containers in the `nerdctl` cli.
@junnplus WDYT

aseaday:
LGTM.  Does nerdctl provide some functions indicate the binary path?
@gaocegege WDYT

junnplus:
> Does nerdctl provide some functions indicate the binary path?

I recommend using `exec.LookPath("nerdctl")` to find.

VoVAllen:
I think this is a good idea, espeically the support of estargz from nerdctl can greatly improve user experiences using large images (most cases in data science scenarios). On the other hand, we may also provide kata-container support later, to deliver a full VM experience(for example user may want to use docker inside container), which is also supported by nerdctl (I guess?). 

However we may need some more investigations about nerdctl to decide how to deliver this.

gaocegege:
I think we should support nerdctl, especially for lima users.

gaocegege:
Lima works like a charm on apple M1. And the container runtime CLI is nerdctl by default.

PR #1378, opened 2023-01-04
feat: :sparkles: init nerdctl support

zwpaper:
Signed-off-by: Wei Zhang <kweizh@gmail.com>

1. upstream buildkit support for nerdctl https://github.com/moby/buildkit/pull/3453
2. use `os/exec` for nerdctl support
3. working on encapsulating docker and nerdctl inside `pkg/driver` https://github.com/zwpaper/envd/blob/5d7660a7727db414d304e431bd1f46384b1d5546/pkg/driver/client.go

fix https://github.com/tensorchord/envd/issues/576


gaocegege:

Please assign me to review if it is ready!

zwpaper:
Hi @gaocegege, this PR is currently done and could use nerdctl to build envs, I will raise another PRs for `up` envs and others.

but we still need to wait for buildkit to release a new version after this PR gets merged https://github.com/moby/buildkit/pull/3472, so that we can use a buildkitd with nerdctl support

zwpaper:
Hi @gaocegege this should be ready for review.

before using nerdctl to build envs, please make sure:
1. nerdctl is inside your PATH, for lima case, please link nerdctl to nerdctl.lima
2. use buildkit image built from master, `zwpaper/buildkit:60e82c` if trust me and don't want to built it yourself

gaocegege:
Thanks!

gaocegege:
I think we could push the buildkit master image to tensorchord/buildkit, or use moby/buildkit:latest, WDYT?

gaocegege (commented):
This PR is to add a new builder type nerd-container, right? Currently we have to use docker to run the image built in nerd buildkitd container.

/cc @kemingy @VoVAllen

zwpaper:
there is an image `moby/buildkit:master`, it was built after I update the PR, and it should be ready to use now.

this PR adds a builder for nerdctl so that we can:

1. add a context option for nerdctl
2. start the buildkitd container by nerdctl
3. use nerdctl to send the build request to buildkitd and then build the env

zwpaper:
for dev and review cases, we could use the master image. when releasing, we should wait for the versioned image for buildkit

pkg/builder/build.go
```diff
@@ -206,7 +207,7 @@ func (b generalBuilder) build(ctx context.Context, pw progresswriter.Writer) err
 
 	for _, entry := range b.entries {
 		// Set up docker config auth.
-		attachable := []session.Attachable{authprovider.NewDockerAuthProvider(os.Stderr)}
+		attachable := []session.Attachable{authprovider.NewDockerAuthProvider(&configfile.ConfigFile{})}
```
gaocegege, line 210: What does this change mean?

zwpaper: the `authprovider` is upgraded with the `buildkit` client library, the NewDockerAuthProvider param is updated from upstream so that I have to change the args

gaocegege: Gotcha.

pkg/driver/nerdctl/nerdctl.go
```diff
@@ -0,0 +1,232 @@
...
+}
+
+func NewClient(ctx context.Context) (driver.Client, error) {
+	bin, err := exec.LookPath("nerdctl")
+	if err != nil {
+		// for mac lima users, check nerdctl.lima
+		bin, err = exec.LookPath("nerdctl.lima")
+		if err != nil {
+			return nil, err
+		}
```
kemingy, line 44: Seems this will override the `err`. For non (mac + lima) user, they will get an `exec: "nerdctl.lima": executable file not found in $PATH`, but actually they need `nerdctl` instead of `nerdctl.lima`.

zwpaper: Hi @kemingy, thanks for pointing this out, I have updated the error to `errors.New("can not found nerdctl(or nerdctl.lima for mac) in PATH")`

kemingy (approved):

zwpaper:
@gaocegege need one more approval, what's your opinion?

PR #1393, opened 2023-01-09
fix: mv rlang test to doc_extra test

kemingy:

R Lang test will cause the e2e-language test timeout. We can move it to the doc test since it's a tier 2 supported language.


gaocegege (approved):
LGTM

Issue #1398, opened 2023-01-10
bug: ERROR exporting to oci image format when directory name contains whitespace


xxchan:
## Additional Info

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritise the issues with the most 👍.

gaocegege:
Thanks for raising this!

PR #1406, opened 2023-01-11
fix: create a legal container env name from cur dir

kemingy:

* fix https://github.com/tensorchord/envd/issues/1398


gaocegege (approved):

PR #1416, opened 2023-01-16
feat: file sync

AlexXi19:
# File sync 
Syncs the user's local project directory to a pod in the kubernetes cluster started by `envd run` with `envd-server`. 

### Steps 
- Installs syncthing binary on the user's device
- Starts local syncthing process 
- Connects to remote syncthing process in the pod 
  - Does so by port forwarding the syncthing api 
- Connects the local device and the remote device by applying a config
  - Local and remote forward of syncthing discovery ports 
- Adds syncing folders by applying a config 

### Waiting for config to apply
Applying configurations through the syncthing rest api is async so we use the [event api](https://docs.syncthing.net/dev/events.html) and wait until we see a config apply success event 

### What's not included and known limitations 
- Currently can only work for one project at a time because all syncthing ports are hardcoded and not fetched using a `getFreePorts`. This change should be pretty easy 

### Syncthing process management 
- The process will exit on `ctrl-c` and `ctrl-d`

### What's next 
- Sync for multiple projects 
- `envd attach` for starting up the sync and port forwarding again

### Testing 
Would like some opinion on what's are the final tests to keep in the repo. 
- Manual walking through the process and seeing if it works.
- Miscellaneous testing written during development to progress check on intermediate functionalities.


gaocegege:

Is this ready to review?

VoVAllen:
@gaocegege it's ready

VoVAllen:
Things to be discussed:
- How user can specify the behavior of code folder, sync vs git clone? Through CLI or envd file?

AlexXi19:
> Things to be discussed:
> 
> * How user can specify the behavior of code folder, sync vs git clone? Through CLI or envd file?

In my opinion, I think sync can replace git clone and there can be an option in `build.envd` or a `run` flag to opt in or out of file syncing.

pkg/app/destroy.go
```diff
@@ -96,5 +98,12 @@ func destroy(clicontext *cli.Context) error {
 		logrus.Infof("failed to remove entry %s from your SSH config file: %s", ctrName, err)
 		return errors.Wrap(err, "failed to remove entry from your SSH config file")
 	}
+
+	homeDirectory := syncthing.GetHomeDirectory(name)
+	err = os.RemoveAll(homeDirectory)
+	if err != nil {
+		logrus.Error(err)
+	}
```
kemingy, line 106: Is it related to CleanLocalConfig?

pkg/syncthing/client.go
```diff
@@ -0,0 +1,77 @@
...
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"net/http"
+	"path"
+	"time"
+)
+
+func NewApiClient() *http.Client {
```
kemingy, line 26:
```suggestion
func NewAPIClient() *http.Client {
```

Follow the naming convention in this project. Also need to fix others.

pkg/app/run.go
```diff
@@ -199,6 +203,57 @@ func run(clicontext *cli.Context) error {
 			}()
 		}
 
+		go func() {
```
kemingy, line 206: This part can be put into another function. It's too long.

pkg/syncthing/client.go
```diff
@@ -0,0 +1,77 @@
...
+	GET  = "GET"
+	POST = "POST"
+	PUT  = "PUT"
+)
+
+// Makes API calls to the syncthing instance's rest api
+func (s *Syncthing) ApiCall(method string, url string, params map[string]string, body []byte) ([]byte, error) {
+	// TODO: can implement retry logic
+
+	var urlPath = path.Join(fmt.Sprintf("localhost:%s", s.Port), url)
```
kemingy, line 42: You cannot join URLs with path join. Its behavior depends on the OS.

pkg/syncthing/client.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+const (
+	GET  = "GET"
+	POST = "POST"
+	PUT  = "PUT"
+)
+
+// Makes API calls to the syncthing instance's rest api
+func (s *Syncthing) ApiCall(method string, url string, params map[string]string, body []byte) ([]byte, error) {
```
kemingy, line 39: This abstraction is not suitable. You can refer to the implementation in https://github.com/tensorchord/envd-server/blob/main/client/client.go#L25

kemingy (commented):
Only reviewed part of it. @gaocegege please take a look when you're free.

pkg/syncthing/file.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/tensorchord/envd/pkg/util/fileutil"
+)
+
+func GetHomeDirectory(name string) string {
+	return fmt.Sprintf("%s/syncthing-%s", fileutil.DefaultConfigDir, name)
+}
+
+func GetConfigFilePath(homeDirectory string) string {
+	return fmt.Sprintf("%s/config.xml", homeDirectory)
+}
```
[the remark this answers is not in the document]
AlexXi19: Yes, i will use this to stay consistent

pkg/app/destroy.go
AlexXi19: clean local config is based on a syncthing struct, which is not present in the destroy command (technically i can construct it and then use CleanLocalConfig). Or just write another function that concise all of this.

pkg/syncthing/client.go
AlexXi19: Ok, i will pay more attention. Thanks for pointing it out.

pkg/app/run.go
AlexXi19: sounds good, i will abstract it

pkg/syncthing/client.go
AlexXi19: 🤔i didnt know that, i will look more into the behavior, thanks for pointing this out

pkg/syncthing/client.go
AlexXi19: Ok, let me take a closer look

pkg/syncthing/util.go
```diff
@@ -0,0 +1,27 @@
...
+	"strings"
+)
+
+func ParsePortFromAddress(addr string) string {
+	if strings.Contains(addr, ":") {
+		lst := strings.Split(addr, ":")
+		return lst[len(lst)-1]
+	}
+	return ""
+}
```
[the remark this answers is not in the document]
AlexXi19: can't use this because the url provided is does not have the a prefix e.g. "http"

AlexXi19:
addressed comments 

cc @kemingy

VoVAllen (changes requested):
Please also add syncthing option at the envd-server side
- https://github.com/tensorchord/envd-server/blob/main/api/types/environment.go#L37 Add a new option here
- Pass the option from client at https://github.com/tensorchord/envd/blob/aab7f53d2cf57549598c73040d98165fb2ecf5cb/pkg/envd/envdserver.go#L281
- Add syncthing option in EnvdServerSource https://github.com/tensorchord/envd/blob/56d25a01dd37b2c069ae9b4492811c8571a1e714/pkg/envd/types.go#L57
- Pass syncthing flag through https://github.com/tensorchord/envd/blob/d8e16a10a9c33979a9402f7e701deca6036befd1/pkg/app/run.go#L137

And also modify the [server part](https://github.com/tensorchord/envd-server/pull/170/files#diff-ff61e7191f455a6edd54f1b34d7dc3384bab2476d8c0889f0db3d4ad80a26694R99), to ensure the git clone logic is not affected is syncthing option is not specified

VoVAllen:
And also please resolve the conflicts. Thanks!

pkg/syncthing/client.go
```diff
@@ -0,0 +1,100 @@
...
+	ApiKey   string
+	Client   *http.Client
+	BasePath string
+}
+
+func (s *Syncthing) NewClient() *Client {
+	return &Client{
+		ApiKey:   s.ApiKey,
+		Client:   NewAPIClient(),
+		BasePath: fmt.Sprintf("http://localhost:%s", s.Port),
```
kemingy, line 49: localhost can be resolved to an IPv6 address. Does this tool support that? Or you just mean the loopback address.

pkg/syncthing/client.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+}
+
+func (c *Client) get(url string, params map[string]string) ([]byte, error) {
+	return c.sendRequest(GET, url, params, nil)
+}
+
+func (c *Client) put(url string, params map[string]string, body []byte) ([]byte, error) {
+	return c.sendRequest(PUT, url, params, body)
+}
```
kemingy, line 59: Looks redundant

pkg/syncthing/syncthing.go
```diff
@@ -0,0 +1,233 @@
...
+	latestEventId int64
+	DeviceAddress string
+}
+
+// Initializes the remote syncthing instance
+func InitializeRemoteSyncthing() (*Syncthing, error) {
+	s := &Syncthing{
+		Name:          "Remote Syncthing",
+		Port:          DefaultRemotePort,
+		HomeDirectory: "/config",
```
kemingy, line 49: Any special reason to use this uncommon path?

pkg/syncthing/install.go
```diff
@@ -0,0 +1,142 @@
...
+	return "1.22.2"
+}
+
+func getSyncthingInstallPath() string {
+	return fmt.Sprintf("%s/bin", fileutil.DefaultCacheDir)
+}
+
+func GetSyncthingBinPath() string {
+	return fmt.Sprintf("%s/syncthing", getSyncthingInstallPath())
+}
```
kemingy, line 40: use path join

pkg/syncthing/syncthing.go
AlexXi19: This is the default for the syncthing image.

pkg/syncthing/client.go
AlexXi19: Will remove

pkg/app/run.go
```diff
@@ -199,6 +208,36 @@ func run(clicontext *cli.Context) error {
 			}()
 		}
 
+		if clicontext.Bool("sync") {
+			go func() {
+				syncthingApiAddr := fmt.Sprintf("127.0.0.1:%s", syncthing.DefaultRemotePort)
```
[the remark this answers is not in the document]
AlexXi19: Changed to constant

pkg/syncthing/install.go
AlexXi19: OK

AlexXi19:
@VoVAllen waiting on `envd-server` release to get latest types.

gaocegege:
I will create a new release today.

AlexXi19:
@VoVAllen updated dependencies, now its all passing :)

gaocegege (commented):
LGTM

Could we merge this?

/cc @kemingy @VoVAllen

kemingy:
> LGTM
> 
> Could we merge this?
> 
> /cc @kemingy @VoVAllen

LGTM

VoVAllen (approved):

PR #1425, opened 2023-01-16
LLM inference example

VoVAllen:

### DIFF
--- a/examples/llm-inference/build.envd
+++ b/examples/llm-inference/build.envd
@@ -1,6 +1,6 @@
 # syntax=v1
 def build():
-    # base(os="ubuntu20.04", language="python3")    
+    # base(os="ubuntu20.04", language="python3")
     base(dev=True)
     install.conda()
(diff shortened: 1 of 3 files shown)


examples/llm-inference/build.envd
```diff
@@ -0,0 +1,18 @@
...
+        [
+            "torch",
+            "torchvision",
+            "bitsandbytes",
+            "accelerate",
+            "--extra-index-url https://download.pytorch.org/whl/cu113",
+        ]
+    )
+
+    install.python_packages(["git+https://github.com/huggingface/transformers.git"])
```
gaocegege, line 18: Why should we install the transformers via git?

kemingy: Tested. We can install it from PyPI directly. No need to use the latest code.

gaocegege: Then could we remove it?

PR #1426, opened 2023-01-16
docs: v1 r_lang and julia packages

kemingy:

One more thing, only `install.julia_packages` use `names` instead of `name`. Shall we align them? cc @VoVAllen @gaocegege

### DIFF
--- a/e2e/v1/docs/testdata/julia/build.envd
+++ b/e2e/v1/docs/testdata/julia/build.envd
@@ -4,5 +4,5 @@
 def build():
     base(dev=True)
     install.julia()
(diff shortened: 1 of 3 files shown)


gaocegege:
I think we should.

kemingy:
> I think we should.

Do you mean changing to `names` or `name` in v1?

gaocegege:
> > I think we should.
> 
> Do you mean changing to `names` or `name` in v1?

Yep

gaocegege:
names

kemingy:
> names could be name

Fixed. PTAL.

gaocegege (approved):

PR #1434, opened 2023-01-21
fix: pip install requirements.txt

kemingy:

Switching users would be complex. It needs re-set all the environment variables.

The old version also takes longer to start installing (tested on WSL, it takes more than 15s to start).

### DIFF
--- a/pkg/lang/ir/v0/python.go
+++ b/pkg/lang/ir/v0/python.go
@@ -157,25 +157,12 @@ func (g generalGraph) compilePyPIPackages(root llb.State) llb.State {
 	}
 
 	if g.RequirementsFile != nil {
-		// Compose the package install command.
-		var sb strings.Builder
-		sb.WriteString("bash -c '")
-		sb.WriteString("set -euo pipefail\n")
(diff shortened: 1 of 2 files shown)


pkg/lang/ir/v0/python.go
```diff
@@ -157,25 +157,12 @@ func (g generalGraph) compilePyPIPackages(root llb.State) llb.State {
...
-		cmd := sb.String()
-
-		logrus.WithField("command", cmd).
+		logrus.WithField("file", *g.RequirementsFile).
 			Debug("Configure pip install requirements statements")
-		root = root.User("root").Dir(g.getWorkingDir())
+		root = root.Dir(g.getWorkingDir())
 		run := root.
-			Run(llb.Shlex(cmd), llb.WithCustomNamef("pip install %s", *g.RequirementsFile))
+			Run(llb.Shlexf("/opt/conda/envs/envd/bin/python -m pip install -r  %s", *g.RequirementsFile),
```
gaocegege, line 164: Does pip cache work after the change? We used root to keep cache in the /root/.pip

kemingy: This is still the root user.

Issue #1429, opened 2023-01-19
feat: --shm-size 8G to the docker run


silakanveli:
## Description

Torch dataloader gives errors. Need to be able to add more memory.

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

gaocegege:
SGTM.

Thanks for the issue!

PR #1445, opened 2023-01-25
feat: add shm size to start options

kemingy:

* fix https://github.com/tensorchord/envd/issues/1429

### DIFF
--- a/pkg/app/run.go
+++ b/pkg/app/run.go
@@ -77,7 +77,7 @@ var CommandCreate = &cli.Command{
 		&cli.IntFlag{
 			Name:  "shm-size",
 			Usage: "Configure the shared memory size (megabyte)",
-			Value: 0,
+			Value: 2048,
 		},
 		&cli.StringFlag{
 			Name:  "cpu",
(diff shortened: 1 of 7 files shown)


kemingy:
Blocked by https://github.com/tensorchord/envd-server/pull/176

pkg/app/run.go
```diff
@@ -74,6 +74,11 @@ var CommandCreate = &cli.Command{
 			Usage: "Assign the host address for environment ssh acesss server listening",
 			Value: envd.Localhost,
 		},
+		&cli.IntFlag{
+			Name:  "shm-size",
+			Usage: "Configure the shared memory size (megabyte)",
+			Value: 0,
```
gaocegege, line 80: I think we should set a larger default value, WDYT?

kemingy: How much? It also consumes memory. The default resource limit is 2GB of memory. 
- https://github.com/tensorchord/envd-server/blob/f9e9d2689a22edda828be867dde626eb188c6824/pkg/runtime/kubernetes/environment_create.go#LL176C30-L176C30

kemingy: Be default, the container will have 64MB of shared memory.

gaocegege: I think for docker, we can keep 2GB memory. WDYT

kemingy: SGTM

gaocegege (approved):
/lgtm

PR #1447, opened 2023-01-25
fix: rm remote cache for v1

kemingy:

We don't have the remote cache for v1. It only slows down the 1st build.

### DIFF
--- a/pkg/lang/ir/v1/compile.go
+++ b/pkg/lang/ir/v1/compile.go
@@ -25,16 +25,13 @@ import (
 	"github.com/cockroachdb/errors"
 	"github.com/moby/buildkit/client/llb"
 	"github.com/sirupsen/logrus"
-	"github.com/spf13/viper"
 	servertypes "github.com/tensorchord/envd-server/api/types"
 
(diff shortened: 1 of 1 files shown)


pkg/lang/ir/v1/compile.go
```diff
@@ -256,20 +253,8 @@ func (g generalGraph) EnvString() []string {
...
-			version.GetVersionForImageTag(), *g.CUDA, g.CUDNN)
-	} else {
-		res = fmt.Sprintf(
-			"type=registry,ref=docker.io/%s/python-cache:envd-%s",
-			viper.GetString(flag.FlagDockerOrganization),
-			version.GetVersionForImageTag())
-	}
-	return &res, nil
+	// We don't have remote cache for v1
+	return nil, nil
```
gaocegege, line 257: LGTM, but should we also keep cache for v1?

kemingy: V1 changes frequently. The remote cache may not work as expected. I think we only need to maintain the remote cache for the stable version.

gaocegege: SGTM

PR #1470, opened 2023-02-08
fix: pass loop variable as a function parameter

hezhizhen:


hezhizhen:
It's a naive way to handle it: wrap the whole part in an anonymous function and pass the loop variable as a function parameter. I think there should be a better to fix it.

Btw, I wonder why the `case` works but the `default` fails.

gaocegege (approved):

PR #1480, opened 2023-02-16
fix: add revive and address indent-error-flow issues

hezhizhen:
Add `revive` linter and enable `indent-error-flow` rule only.


gaocegege:

PR #1482, opened 2023-02-16
feat: Allow buildkit timeout to be configurable

terrytangyuan:
This was needed for me to get everything to run. I'd imagine other users might benefit from this as well.

### DIFF
--- a/pkg/util/envutil/env.go
+++ b/pkg/util/envutil/env.go
@@ -21,6 +21,10 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+// GetDurationWithDefault parses a duration string from environment variable with the given key
+// or uses the given default duration. The duration string is a possibly signed sequence of
+// decimal numbers, each with optional fraction and a unit suffix, such as "300ms", "-1.5h" or "2h45m".
+// Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
 func GetDurationWithDefault(key string, o time.Duration) time.Duration {
 	v, found := os.LookupEnv(key)
 	if found && v != "" {
--- a/pkg/util/envutil/env.go
+++ b/pkg/util/envutil/env.go
@@ -1,0 +1,34 @@
+// Copyright 2023 The envd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package envutil
+
+import (
+	"os"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+)
+
+func GetDurationWithDefault(key string, o time.Duration) time.Duration {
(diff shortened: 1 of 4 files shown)


pkg/util/envutil/env_test.go
```diff
@@ -0,0 +1,34 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestLookupEnvDurationOr(t *testing.T) {
+	defer func() { _ = os.Unsetenv("FOO") }()
+	assert.Equal(t, time.Second, LookupEnvDurationOr("", time.Second), "default value")
+	_ = os.Setenv("FOO", "bar")
```
hezhizhen, line 28: use `t.Setenv()` in tests.

And the `defer` will be no longer needed.

pkg/util/envutil/env.go
```diff
@@ -0,0 +1,35 @@
...
+package envutil
+
+import (
+	"os"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+)
+
+func LookupEnvDurationOr(key string, o time.Duration) time.Duration {
```
hezhizhen, line 24: How about `GetDurationWithDefault`? I think there is no need to emphasize the action `lookup env`.

pkg/util/envutil/env.go
```diff
@@ -0,0 +1,35 @@
...
+	v, found := os.LookupEnv(key)
+	if found && v != "" {
+		d, err := time.ParseDuration(v)
+		if err != nil {
+			log.WithField(key, v).WithError(err).Panic("failed to parse")
+		} else {
```
hezhizhen, line 30: Handle this case first because it returns directly.

terrytangyuan:
@hezhizhen Great suggestions! Updated.

pkg/util/envutil/env.go
```diff
@@ -0,0 +1,34 @@
...
+package envutil
+
+import (
+	"os"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+)
+
+func GetDurationWithDefault(key string, o time.Duration) time.Duration {
```
hezhizhen, line 24: It's better to give some explanation about accepted time syntax in comments.

pkg/util/envutil/env_test.go
```diff
@@ -0,0 +1,32 @@
...
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetDurationWithDefault(t *testing.T) {
+	assert.Equal(t, time.Second, GetDurationWithDefault("", time.Second), "default value")
+	t.Setenv("FOO", "bar")
+	assert.Panics(t, func() { GetDurationWithDefault("FOO", time.Second) }, "bad value")
+	t.Setenv("FOO", "1h")
```
hezhizhen, line 28: How about using the actual environment variable name?

pkg/util/envutil/env.go
```diff
@@ -0,0 +1,34 @@
...
+	v, found := os.LookupEnv(key)
+	if found && v != "" {
+		d, err := time.ParseDuration(v)
+		if err == nil {
+			return d
+		}
+		log.WithField(key, v).WithError(err).Panic("failed to parse")
```
hezhizhen, line 31: will `Panic()` panic here?

terrytangyuan: Yes it's intentional since we don't want users to assume that the incorrect config is applied when actually not.

gaocegege:
PTAL @hezhizhen

hezhizhen:
LGTM

gaocegege:

And Thank @hezhizhen for the review.

PR #1484, opened 2023-02-17
Julia MNIST classification e2e test example

oubotong:
Here is the example of Julia doing a classification task on the MNIST dataset using traditional CNN.
It will print out an average accuracy on the test dataset.


terrytangyuan (approved):

kemingy (approved):
LGTM

PR #1486, opened 2023-02-17
feat: support CPU and memory limit for docker runner

kemingy:
cc @Zheaoli

### DIFF
--- a/pkg/app/up.go
+++ b/pkg/app/up.go
@@ -90,6 +90,16 @@ var CommandUp = &cli.Command{
 			Usage: "Configure the shared memory size (megabyte)",
 			Value: 2048,
 		},
+		&cli.StringFlag{
+			Name:  "cpu",
+			Usage: "Request CPU resources (number of cores), such as 0.5, 1, 2",
+			Value: "",
+		},
+		&cli.StringFlag{
+			Name:  "memory",
+			Usage: "Request Memory, such as 512M, 2G",
+			Value: "",
+		},
 		&cli.BoolFlag{
 			Name:  "detach",
 			Usage: "Detach from the container",
@@ -205,6 +215,8 @@ func up(clicontext *cli.Context) error {
 		Timeout:         clicontext.Duration("timeout"),
 		SshdHost:        clicontext.String("host"),
 		ShmSize:         clicontext.Int("shm-size"),
+		NumCPU:          clicontext.String("cpu"),
+		NumMem:          clicontext.String("memroy"),
 	}
 	if c.Runner != types.RunnerTypeEnvdServer {
 		startOptions.EngineSource = envd.EngineSource{
--- a/pkg/app/up.go
+++ b/pkg/app/up.go
@@ -97,7 +97,7 @@ var CommandUp = &cli.Command{
 		},
 		&cli.StringFlag{
 			Name:  "memory",
-			Usage: "Request Memory, such as 512M, 2G",
+			Usage: "Request Memory, such as 512Mb, 2Gb",
 			Value: "",
 		},
 		&cli.BoolFlag{
@@ -216,7 +216,7 @@ func up(clicontext *cli.Context) error {
 		SshdHost:        clicontext.String("host"),
 		ShmSize:         clicontext.Int("shm-size"),
 		NumCPU:          clicontext.String("cpu"),
-		NumMem:          clicontext.String("memroy"),
+		NumMem:          clicontext.String("memory"),
 	}
 	if c.Runner != types.RunnerTypeEnvdServer {
 		startOptions.EngineSource = envd.EngineSource{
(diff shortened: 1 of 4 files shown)


pkg/envd/docker.go
```diff
@@ -459,6 +463,23 @@ func (e dockerEngine) StartEnvd(ctx context.Context, so StartOptions) (*StartRes
 	if so.ShmSize > 0 {
 		hostConfig.ShmSize = int64(so.ShmSize) * 1024 * 1024
 	}
+	// resource
+	if len(so.NumCPU) > 0 {
+		cpu, err := strconv.ParseFloat(so.NumCPU, 64)
+		if err != nil {
+			logger.Infof("parse `cpu` error: %v, ignore this argument", err)
+		} else {
+			hostConfig.NanoCPUs = int64(cpu * 10e9)
```
Zheaoli, line 472: I prefer not to use the `NanoCPUs`. I think  I would be better to set the  CPU quota by using `CPUPeriod` and `CPUQuota`. I think it would be better to get less CPU throttle.

pkg/app/up.go
```diff
@@ -90,6 +90,16 @@ var CommandUp = &cli.Command{
 			Usage: "Configure the shared memory size (megabyte)",
 			Value: 2048,
 		},
+		&cli.StringFlag{
```
Zheaoli, line 93: should we add another flag named `enable-cpu-binding` to make people can decide to bind CPU core nor not

kemingy: Do you mean `--cpuset-cpus` for docker?

pkg/envd/docker.go
kemingy: The main reason I choose `NanoCPUs` is because it's applicable to all platforms. But I guess UNIX is also okay. WDYT cc @gaocegege

kemingy: Or we can detect the platform and apply `CPUPeriod` for UNIX.

Zheaoli: > Or we can detect the platform and apply `CPUPeriod` for UNIX.

SGTM

pkg/app/up.go
Zheaoli: > Do you mean `--cpuset-cpus` for docker?

Yes, the `cpuset-cpus` means binding the workload with CPU core to reach the max performance

kemingy:
Comments addressed. PTAL.

pkg/app/up.go
```diff
@@ -90,6 +90,21 @@ var CommandUp = &cli.Command{
...
+			Name:  "cpu",
```
suyanhanx, line 94:
```sugeestion
    Name:  "cpus",
```

nit: naming "cpus" for the number of CPUs

pkg/envd/docker.go
```diff
@@ -459,6 +465,30 @@ func (e dockerEngine) StartEnvd(ctx context.Context, so StartOptions) (*StartRes
...
+		} else if runtime.GOOS == "windows" {
+			hostConfig.NanoCPUs = int64(cpu * 10e9)
+		} else {
+			// refer to https://docs.docker.com/config/containers/resource_constraints/#configure-the-default-cfs-scheduler
+			// CPU quota and CPU period only work for UNIX platform
+			hostConfig.CPUQuota = int64(cpu * 10e5)
+		}
+	}
+	if len(so.CPUSet) > 0 {
+		hostConfig.CpusetCpus = so.CPUSet
```
Zheaoli, line 482: the cpuset argument and numcpu should be exited one at the same time.

Maybe we can do something like this

1. use a boolean flag named `use-cpu-set` or something else.
2. if the people set the `use-cpu-set` , we can do something below
    1. if the `cpu-set` is not set, we can calculate the CPU core we want to binding.
        1. if `cpu` is a float, we can make it as a integer, such as 0.5 -> 1, 1.5 -> 2. And we can raise a warning during the code execution
        2. if `cpu` is a integer, we can use it directly. such as 1 cpu binding to 0, 2 cpu binding to 0,1 etc.
    2. if the `cpu-set` is set, we can use it directly(`cpu` should be ignored and warning it during in execution)
3. if the people don't set the `use-cpu-set`, we can use the `cpu` directly(raise a warning message if people set `cpu-set`).

Zheaoli: cc @gaocegege

gaocegege: SGTM!

pkg/app/up.go
kemingy: fixed

pkg/envd/docker.go
kemingy: I think converting `cpu` to `cpu-set` when `use-cpu-set` is enabled is tricky:
1. round the CPU quota may not be something expected
2. bind to cpu-set requires storing the previous environment information, otherwise we don't know which one is still free to use
3. this may only work well when the server is fully used for envd environments

Thus I think we can just panic if both of them are set.

Zheaoli (approved):
LGTM

suyanhanx (approved):
LGTM.

gaocegege:
Thanks for your review 🥰

PR #1495, opened 2023-02-20
feat: Support set environment name in `envd up` command

Zheaoli:


aseaday:
Is `StringFlag`'s default value to `""` defined or undefined which may break someday in the future?

aseaday:
Please check my comment and other parts LGTM.

kemingy:
Actually, `-t` is the environment name.

Zheaoli:
> Actually, `-t` is the environment name.

I think `-t` is not working for env name. FYI https://github.com/tensorchord/envd/blob/main/pkg/app/up.go#L210

kemingy:
But I think there is a bug for `-t`. If you also set it to `envd up -t test:latest` then you cannot destroy it because `envd destroy --name xxx` doesn't accept the tag. :(

Zheaoli:
All comment has been addressed, PTAL

aseaday:
> But I think there is a bug for `-t`. If you also set it to `envd up -t test:latest` then you cannot destroy it because `envd destroy --name xxx` doesn't accept the tag. :(

Good question. I also have the confusion among the tag, image and the env

suyanhanx:
It looks confusing. Other commands use another different flag to indicate the name.
https://github.com/tensorchord/envd/blob/a4f7a5ca824d2ce375b01fb61bfe3c1e48aef3b2/pkg/app/pause.go#L26-L35

I think "name" would be a better option as a flag to indicate the name. 
The "-t"/tag seems more appropriate for images.

Or to put it another way, the concept of "tag" does not seem to be appropriate for defining environments. `test:latest` can already be a complete name, can't it?

kemingy:
> I think "name" would be a better option as a flag to indicate the name. The "-t"/tag seems more appropriate for images.
> 
> Or to put it another way, the concept of "tag" does not seem to be appropriate for defining environments. `test:latest` can already be a complete name, can't it?

Agree. I think we need to align the naming.

Zheaoli:
so, is there any final decision about this PR?

kemingy:
cc @gaocegege

kemingy (approved):

kemingy:
I misunderstood. It's the `environment` name, not the image name. The code has so many *name* things.

PR #1500, opened 2023-02-22
build: Reduce redundant output

tisonkun:
Follow suggestions in https://github.com/tensorchord/envd/discussions/1498#discussioncomment-5074168.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,7 +102,7 @@ export GOFLAGS ?= -count=1
 
 build-release:
 	@for target in $(TARGETS); do                                                      \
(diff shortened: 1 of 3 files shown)


tisonkun:
Interesting. It seems the test depends on local config env. I'll revert the test-related changes but perhaps further investigation for stably run tests are helpful.

Said that current context initialization depends on `$HOME/.config/envd/context` and it can be polluted by other commands.

gaocegege:

tisonkun:
@gaocegege I just notice that there are `build-release` _and_ `build-local`...I'll submit another patch and check all build command.

Issue #1563, opened 2023-04-20
bug: rust and extension are installed in wrong path


n063h:
### Are you use the envd server?

- [ ] Yes, I am using the envd server.
- [X] No, I am not using the envd server.

### Describe the bug

when I installed rust and vscode_extension "rust-lang.rust-analyzer", I found the softwares are installed in `/root` instead of `/home/envd`. This is not what it should be.

```bash
⬢ [envd]❯ ls ~ -a
.  ..  .bash_logout  .bashrc  .cache  .config  .profile  .vscode-server
⬢ [envd]❯ sudo ls /root -a
.  ..  .bashrc  .cargo  .profile  .rustup  .zshenv
```

Thus, when I run `cargo`, error occurs `bash: cargo: command not found`

### To Reproduce

envd up with 
```python
# syntax=v1

def rust():
    """Install Rust."""
    install.apt_packages(name=["build-essential"])
    run(
        [
            "curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y",
        ]
    )
    runtime.environ(extra_path=["/home/envd/.cargo/bin"])
    
def build():
    base(dev=True)
    rust()
    install.vscode_extensions(["rust-lang.rust-analyzer"])
    runtime.init(["make install"])
```
In container, run `ls ~ -a`,

### Expected behavior

rust can be installed in /home/envd correctly.
run `ls /home/envd -a`, can see
```bash
⬢ [envd]❯ ls /home/envd -a
.   . bash_logout  . cache  . config  . profile  . sudo_as_admin_successful  . zshenv
..  . bashrc       . cargo  . rustup   . vscode-server
```

### The `docker info` output

Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.9.1)
  compose: Docker Compose (Docker Inc., v2.12.2)
  dev: Docker Dev Environments (Docker Inc., v0.0.3)
  extension: Manages Docker extensions (Docker Inc., v0.2.13)
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc., 0.6.0)
  scan: Docker Scan (Docker Inc., v0.21.0)

Server:
 Containers: 1
  Running: 1
  Paused: 0
  Stopped: 0
 Images: 6
 Server Version: 20.10.21
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
[...]

PR #1564, opened 2023-04-20
fix: compile vscode extensions with user state

n063h:
refer to #1563

# description
Currently, the vscode extensions are installed from `llb.Scratch()`, so some softwares could be installed in /root instead of user envd homeDir.

envd up with
```
# syntax=v1

def rust():
    """Install Rust."""
    install.apt_packages(name=["build-essential"])
    run(
        [
            "curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y",
        ]
    )
    runtime.environ(extra_path=["/home/envd/.cargo/bin"])
    
def build():
    base(dev=True)
    rust()
    install.vscode_extensions(["rust-lang.rust-analyzer"])
    runtime.init(["make install"])
```

# Before
In container,
```bash
⬢ [envd]❯ ls ~ -a
.  ..  .bash_logout  .bashrc  .cache  .config  mosec  .profile  .vscode-server

⬢ [envd]❯ sudo ls /root -a
.  ..  .bashrc  .cargo  .profile  .rustup  .zshenv
```
.cargo and .rustup are installed in wrong path `/root`.

# After
```bash
⬢ [envd]❯ ls ~ -a
.   . bash_logout  . cache  . config  . profile  . sudo_as_admin_successful  . zshenv
..  . bashrc       . cargo  mosec    . rustup   . vscode-server

⬢ [envd]❯ sudo ls /root/ -a
.  ..  . bashrc  . profile
```
Now .cargo and .rustup are installed in correct path `~`.

### DIFF
--- a/pkg/lang/ir/v1/compile.go
+++ b/pkg/lang/ir/v1/compile.go
@@ -336,14 +336,11 @@ func (g *generalGraph) CompileLLB(uid, gid int) (llb.State, error) {
 		if err != nil {
 			return llb.State{}, errors.Wrap(err, "failed to compile entrypoint")
 		}
-		vscode, err := g.compileVSCode()
+		vscode, err := g.compileVSCode(entrypoint)
 		if err != nil {
 			return llb.State{}, errors.Wrap(err, "failed to compile VSCode extensions")
 		}
-		copy = llb.Merge([]llb.State{
-			entrypoint,
-			vscode,
-		}, llb.WithCustomName("[internal] final dev environment"))
+		copy = vscode
 	}
 
 	// it's necessary to exec `run` with the desired user
--- a/pkg/lang/ir/v1/editor.go
+++ b/pkg/lang/ir/v1/editor.go
@@ -31,9 +31,9 @@ import (
 
 func (g generalGraph) compileVSCode(root llb.State) (llb.State, error) {
 	if len(g.VSCodePlugins) == 0 {
-		return llb.Scratch(), nil
+		return root, nil
 	}
-	ext := root
+	inputs := []llb.State{}
 	for _, p := range g.VSCodePlugins {
 		vscodeClient, err := vscode.NewClient(vscode.MarketplaceVendorOpenVSX)
 		if err != nil {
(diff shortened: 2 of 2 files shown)


pkg/lang/ir/v1/compile.go
```diff
@@ -336,14 +336,11 @@ func (g *generalGraph) CompileLLB(uid, gid int) (llb.State, error) {
 		if err != nil {
 			return llb.State{}, errors.Wrap(err, "failed to compile entrypoint")
 		}
-		vscode, err := g.compileVSCode()
+		vscode, err := g.compileVSCode(entrypoint)
```
kemingy, line 339: I remember you mention that this should happen before the `run` command.

n063h: Later I found that the issue of rust-analyzer not supporting the current platform is not due to `compileVSCode` first and then compile `run`. The rust-analyzer server checks rust every time it starts up, rather than checking it only once during installation. So it's not a matter.

pkg/lang/ir/v1/editor.go
```diff
@@ -29,11 +29,11 @@ import (
 	"github.com/tensorchord/envd/pkg/util/fileutil"
 )
 
-func (g generalGraph) compileVSCode() (llb.State, error) {
+func (g generalGraph) compileVSCode(root llb.State) (llb.State, error) {
 	if len(g.VSCodePlugins) == 0 {
 		return llb.Scratch(), nil
```
kemingy, line 34: This should return root.

pkg/lang/ir/v1/editor.go
```diff
@@ -45,17 +45,15 @@ func (g generalGraph) compileVSCode() (llb.State, error) {
...
+		ext = ext.File(llb.Copy(llb.Local(flag.FlagCacheDir),
 			vscodeClient.PluginPath(p),
 			fileutil.EnvdHomeDir(".vscode-server", "extensions", p.String()),
 			&llb.CopyInfo{
 				CreateDestPath: true,
 			}, llb.WithUIDGID(g.uid, g.gid)),
 			llb.WithCustomNamef("install vscode plugin %s", p.String()))
-		inputs = append(inputs, ext)
 	}
-	layer := llb.Merge(inputs, llb.WithCustomName("merging plugins for vscode"))
```
kemingy, line 57: Can they run in parallel?

pkg/lang/ir/v1/compile.go
kemingy: So will it be resolved to the correct platform?

n063h: It's a seperate problem, I'll open a new issue and fix it.

pkg/lang/ir/v1/editor.go
n063h: yes

pkg/lang/ir/v1/editor.go
n063h: The next commit should make it possible

kemingy (approved):

Issue #1336, opened 2022-12-23
feat: Better clarity for `envd destroy` 


AlexXi19:
## Description
Using `envd destroy` incorrectly can be a bit confusing. 

Here I was trying to destroy `airsolar` and the response seemed successful. However, `envd_playground` was the one that actually got deleted. This can be a bit confusing because the error message shows that the environment is successfully destroyed (even though it's not the right one). 

![image](https://user-images.githubusercontent.com/68758451/209287909-bbad667a-2db3-45de-bc88-62f48d72faf4.png)

There could be an extra step to determine cases like these where the users are using the destroy command incorrectly and help them notice the error. 

---

**Message from the maintainers**:

Love this enhancement proposal? Give it a 👍. We prioritise the proposals with the most 👍.

kemingy:
I think it's also related to https://github.com/tensorchord/envd/issues/1075

AlexXi19:
![image](https://user-images.githubusercontent.com/68758451/210036449-ff1940f5-1dc2-416c-b6a2-cb74461ea192.png)

To add on to this, deleting non-existent environments can also show success, which can be confusing if you make a typo.

cutecutecat:
If `path` and `name` are not specified either, `envd` will destroy `build.envd` at `cwd` path as default.
This is confiusing, I agree, to let it fail fast here maybe better.

https://github.com/tensorchord/envd/blob/3c38b5e62753d51ec470586b44ada1b14af6eb64/pkg/app/destroy.go#L59

hxu296:
Would #1075 be a blocker for this issue? Otherwise I would like to work on this issue :)

I'm learning this codebase and this issue seems to be a good starting point. @gaocegege  @cutecutecat

gaocegege:
Personally, I do not think it is a blocker, WDYT @kemingy @cutecutecat

kemingy:
It's not blocked but the user experience is affected by #1075.

gaocegege:
Yeah, that's right.

hxu296:
Thanks for the prompt response! I will start working on this issue.

PR #1568, opened 2023-04-22
feat: confirm `envd destroy` when path and name are both empty

hxu296:
## Summary
The previous behavior of `envd destroy` without both `--path` and `--name` flags was to attempt to delete the environment in the current working directory. This could result in accidental deletion of environments, as seen in issue #1336. This pull request prevents this behavior by adding additional user confirmation to `envd destroy` if both flags are not provided.

### Previous Behavior
```
$ envd destroy
INFO[2023-04-22T06:49:10Z] image(sha256:1a891bf790ef45aec292f63a414fcc4338e1b9e911a313570fb592f9e316a50f) is destroyed 
INFO[2023-04-22T06:49:10Z] environment(envd-quick-start) is destroyed  
```

### Current Behavior
```
$ envd destroy
Are you sure you want to destroy container envd in the current directory? [Y/n]
```

Fixes: #1336

<strike>
Additionally, I have raised [another PR](https://github.com/tensorchord/envd-docs/pull/281) in the envd-docs repository to update the `envd destroy` examples in the documentation to reflect this update.
</strike>

### DIFF
--- a/pkg/app/destroy.go
+++ b/pkg/app/destroy.go
@@ -94,7 +94,8 @@ func destroy(clicontext *cli.Context) error {
 			return errors.Wrap(err, "failed to create an env name")
 		}
 	} else {
-		// Destroy the environment in the current directory if user confirms
+		// Both path and name are empty
+		// Destroy the environment in the current directory only if user confirms
 		buildContext, err := filepath.Abs(".")
 		if err != nil {
 			return errors.Wrap(err, "failed to get absolute path of the build context")
--- a/pkg/app/destroy.go
+++ b/pkg/app/destroy.go
@@ -15,9 +15,13 @@
 package app
 
 import (
+	"bufio"
+	"fmt"
+	"os"
 	"path/filepath"
 
 	"github.com/cockroachdb/errors"
+	"github.com/mattn/go-isatty"
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli/v2"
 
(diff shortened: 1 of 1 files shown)


gaocegege:

/assign @kemingy

pkg/app/destroy.go
```diff
@@ -58,7 +58,7 @@ func destroy(clicontext *cli.Context) error {
 		return errors.New("Cannot specify --path and --name at the same time.")
 	}
 	if path == "" && name == "" {
-		path = "."
+		return errors.New("Must specify --path or --name. If you want to destroy the current environment, please run `envd destroy -p .`")
```
kemingy, line 61: This will be a breaking change.

Personally, I prefer to confirm with users (if tty) if neither of them is provided. At least, it's faster than typing the name or path or waiting for shell completion.

BTW, we should print a log if the specific env doesn't exist. But the exit code could still be 0.

hxu296: Thank you for the prompt review. Confirming with user beforing deleteing the environment in `cwd` sounds more convenient while prevents breaking change. I am not sure what's the most standard way to display confirmation to tty. Could you provide me with some pointers? Would [pkg/app/interactive.go](https://github.com/tensorchord/envd/blob/main/pkg/app/interactive.go) be a good reference?

kemingy: Yeah, you can use that one. Or read from `os.Stdin` directly. Both work for me.

hxu296: Thank you for your input! Reading from `os.Stdin` sounds more readable and easy to implement. I'd go with that :)

hxu296:
As we discussed, now if both `name` and `path` are absent, `envd destory` will confirm with user `Are you sure you want to destroy container ${ctrName} in the current directory? [Y/n]` before deleting the container in `cwd`. 

If there's no tty and both `name` and `path` are absent, `envd destroy` will do nothing.

kemingy (approved):

PR #1577, opened 2023-04-25
fix: hard code amd64 to support remote build

gaocegege:
cc @kemingy


kemingy (approved):

gaocegege:
Currently, we do not support arm64.

@kemingy Could we support arm build in envd v1? I remember that up does not work. But not sure if build works.

gaocegege:
@wild-endeavor I created a new issue to keep track. Does it work for you if we introduce a new CLI flag?

pingsutw:
@gaocegege yes, that will be great

Issue #1580, opened 2023-04-26
feat: Support a new CLI flag `--platform`


gaocegege:
### Describe the feature

Currently, we will use the platform ("amd64, aarch64") and hard coded system "linux" to run the envd build and up commands.

It does not work with the envd context well.

For example, if the context is a remote builder `tcp://xxx`, the platform of the builder may be different from the machine running `envd up` or `envd build`.

Thus we should have a new CLI flag to let users specify the platform. If it is not set, we still use the default platform.

### Why do you need this feature?

https://github.com/tensorchord/envd/pull/1577#issuecomment-1522532074

### Additional context

We should revert this commit https://github.com/tensorchord/envd/commit/ad73e37a17a4142c903d9f7bd4434f67708384fa

PR #1582, opened 2023-04-28
feat: build image by platform

n063h:
refer to #1580

# Before
Hardcoded linux/amd64 platform into code, so envd builds image on linux/amd64 platform by default.

# After
support --platform flag to build image on target platform.


pingsutw:
@n063h Thanks for working on this. Any update for it. we need muti-arch image. I can help for testing

n063h:
> @n063h Thanks for working on this. Any update for it. we need muti-arch image. I can help for testing

@pingsutw  Thanks for reaching out! There's a small hiccup with the dependency "horust" image that I'm working on fixing. I'll get it sorted out soon.

n063h:
@pingsutw Hi, sorry for the wait, I think this branch is ready for your test now. You can run `envd build --platform linux/amd64,linux/arm64`  to build multi-platform images.
Since I have no machine on Non-x86 platforms, I tested it with qemu on linux/arm64 which worked well. 
Please note: horust now provides only linux/amd64,linux/arm64 images, so the images built from envd also work just for the two platforms.

kemingy:
Is it ready to review?

BTW the dev env only work for aarch64/amd64, but the non-dev env should work for more platforms.

n063h:
> Is it ready to review?
> 
> BTW the dev env only work for aarch64/amd64, but the non-dev env should work for more platforms.

yes. 
And the default target-platform is "linux/amd64" no matter what the build-platform is, which may require explicitly flag for arch arm64 for macos users.

gaocegege (approved):
/cc @kemingy

pkg/app/build.go
```diff
@@ -106,12 +112,23 @@ func build(clicontext *cli.Context) error {
...
 
-	builder, err := buildutil.GetBuilder(clicontext, opt)
-	if err != nil {
-		return err
-	}
-	if err = buildutil.InterpretEnvdDef(builder); err != nil {
-		return err
+	platforms := strings.Split(opt.Platform, ",")
+	for _, platform := range platforms {
+		o, p := opt, platform
```
kemingy, line 117: May not need to copy the platform here?

pkg/app/build.go
```diff
@@ -106,12 +112,23 @@ func build(clicontext *cli.Context) error {
...
+		o.Platform = p
+		if len(platforms) > 1 {
+			o.Tag += "-" + strings.Replace(p, "/", "-", 1)
+		}
```
kemingy, line 121: What does this do? Can you add some comments?

pkg/app/build.go
```diff
@@ -89,6 +90,11 @@ To build and push the image to a registry:
 			Usage:   "Import the cache (e.g. type=registry,ref=<image>)",
 			Aliases: []string{"ic"},
 		},
+		&cli.StringFlag{
+			Name:        "platform",
+			Usage:       "Specify the target platforms for the build output (for example, windows/amd64 or linux/amd64,darwin/arm64)",
+			DefaultText: "linux/amd64",
```
kemingy, line 96: I wonder if we should use the local host platform as the default one.

pkg/app/build.go
n063h: build images with same tag could cause image overwriting, so here I add an platform suffix to differentiate them.

pkg/app/build.go
n063h: Ok, I will use the GOOS/GOARCH as the default one

pkg/app/build.go
kemingy: This needs to be documented in CLI help. We'd better create a new issue to track it.

kemingy:
LGTM. Please resolve the conflicts.

pingsutw (commented):
could we also update the platform for [NewImageSource](https://github.com/tensorchord/envd/blob/fd45847e4ebec79bf644d291fa4d2420ea032373/pkg/lang/ir/util.go#L35)
```go
sys := types.SystemContext{}
if platform != nil {
  sys.ArchitectureChoice = platform.Architecture
  sys.OSChoice = platform.OS
}
src, err := ref.NewImageSource(ctx, &sys)
```

pingsutw:
It will pull the base image. I'm using envd in my macbook. It still tries to pull darwin/amd64 base image, although I set the platform to linux/amd64, so we also need to update the platform for NewImageSource.
https://github.com/tensorchord/envd/pull/1587

n063h:
> It will pull the base image. I'm using envd in my macbook. It still tries to pull darwin/amd64 base image, although I set the platform to linux/amd64, so we also need to update the platform for NewImageSource. #1587

I got you, the update has been pushed.

pingsutw (approved):
LGTM, thank you

pingsutw:
btw, @kemingy could you help me cut a release after it merges? I need to bump envd version in flytekit.

kemingy (approved):

kemingy:
> btw, @kemingy could you help me cut a release after it merges? I need to bump envd version in flytekit.

https://github.com/tensorchord/envd/releases/tag/v0.3.22 is released

PR #1593, opened 2023-05-06
fix: merge language packages to system packages

kemingy:
Before this PR, Ubuntu:18.04 (which contains python2.7) will use the default system Python instead of the conda Python because the merge op will override the language packages layer.


kemingy:
pending for 1 approve @gaocegege

gaocegege (approved):

Issue #1636, opened 2023-06-05
bug: 401 Unauthorized for private image registries while fetching base image metadata


mucahitkantepe:
### Are you use the envd server?

- [x] Yes, I am using the envd server.
- [ ] No, I am not using the envd server.

### Describe the bug

When using any private image registry, although local docker setup is authenticated, envd build command fails with 401 Unauthorized response. Tested with ECR and Artifactory

### To Reproduce

have an envd file like below 
```
# syntax=v1
def build():
    base(image="your.private.registry/your-private-image:your-tag", dev=False)
```
then run
```
envd build -f build.envd
```

### The `docker info` output

```
Client:
 Version:    24.0.2
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.10.5
    Path:     /Users/xxx.yyy@zzz.com/.docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.18.1
    Path:     /Users/xxx.yyy@zzz.com/.docker/cli-plugins/docker-compose
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.0
    Path:     /Users/xxx.yyy@zzz.com/.docker/cli-plugins/docker-dev
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.19
    Path:     /Users/xxx.yyy@zzz.com/.docker/cli-plugins/docker-extension
  init: Creates Docker-related starter files for your project (Docker Inc.)
    Version:  v0.1.0-beta.4
    Path:     /Users/xxx.yyy@zzz.com/.docker/cli-plugins/docker-init
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc.)
    Version:  0.6.0
    Path:     /Users/xxx.yyy@zzz.com/.docker/cli-plugins/docker-sbom
  scan: Docker Scan (Docker Inc.)
    Version:  v0.26.0
    Path:     /Users/xxx.yyy@zzz.com/.docker/cli-plugins/docker-scan
  scout: Command line tool for Docker Scout (Docker Inc.)
    Version:  v0.12.0
    Path:     /Users/xxx.yyy@zzz.com/.docker/cli-plugins/docker-scout

Server:
 Containers: 3
  Running: 1
  Paused: 0
  Stopped: 2
 Images: 3
 Server Version: 24.0.2
[...]

PR #1638, opened 2023-06-07
Upgrade buildkitd to v0.11.6

mucahitkantepe:


kemingy:
Thanks. I tried this yesterday but it seems this doesn't work. Have you tried this?

mucahitkantepe:
Do you mean it does not fix the issue or something else does not work with this change?

kemingy:
> Do you mean it does not fix the issue or something else does not work with this change?

It doesn't fix the 401 error.

kemingy:
@gaocegege Shall we merge this one? Do you have any concerns about upgrading the buildkit?

Although this PR still cannot fix the https://github.com/tensorchord/envd/issues/1636

gaocegege:
I think we can. But I wanna know how to fix #1636

PR #1666, opened 2023-06-21
added support for private Docker registries

RichhLi:
Usage:

`envd bootstrap --registry-config path_to_json_file`

Format:
Screenshot 2023-06-23 at 9 32 52 PM

Note that the "name" field is required. The other fields are all optional.

This will allow users to specify private Docker registries and the respective credentials. Under the hood, we parse these credentials and create a corresponding buildkit TOML file, as well as set up Docker bindings. This enables users to pull and push images from private Docker registries.


pkg/app/bootstrap.go
```diff
@@ -112,56 +107,70 @@ func bootstrap(clicontext *cli.Context) error {
...
+					exist, err := fileutil.FileExists(kv[1])
+					if err != nil {
+						return errors.Wrap(err, fmt.Sprintf("failed to parse file path %s", part))
+					}
+					if !exist {
+						return fmt.Errorf("file %s doesn't exist", kv[1])
+					}
+
+					// Generate file path
+					path, err := fileutil.ConfigFile(fmt.Sprintf("registry_%s.pem", kv[0]))
```
ByronHsu, line 150: I am curious why do we have to read and write the pem files? can't we just use the original one?

pkg/app/bootstrap.go
```diff
@@ -112,56 +107,70 @@ func bootstrap(clicontext *cli.Context) error {
...
-	if len(mirror) == 0 {
-		return errors.New("`registry-ca-keypair` should be used with `dockerhub-mirror`")
-	}
+	// Loop over all registry strings
+	if clicontext.IsSet("registry") {
+		for _, reg := range strings.Split(clicontext.String("registry"), ";") {
+			// Split into parts
+			parts := strings.Split(reg, ",")
+			if len(parts) != 4 {
+				return fmt.Errorf("invalid registry string: %s", reg)
```
ByronHsu, line 116: maybe use `errors.Newf` everywhere for consistency?

pkg/app/bootstrap.go
```diff
@@ -315,3 +331,29 @@ func buildkit(clicontext *cli.Context) error {
 
 	return nil
 }
+
+func updateConfigWithRegistry(clicontext *cli.Context, config *buildkitutil.BuildkitConfig) error {
```
ByronHsu, line 335: would be better to add a test to this func

pkg/util/buildkitutil/buildkit.go
```diff
@@ -20,21 +20,31 @@ import (
...
-  {{ if .SetCA}}ca=["/etc/registry/ca.pem"]
-  [[registry."{{ if .Registry }}{{ .Registry }}{{ else }}docker.io{{ end }}".keypair]]
-	key="/etc/registry/key.pem"
-	cert="/etc/registry/cert.pem"
-  {{ end }}
+[registry]
+{{- range $index, $value := .Registry }}
+  [registry."{{ if $value }}{{ $value }}{{ else }}docker.io{{ end }}"]
+  	{{- if $.UseHTTP }}
+    http = true
```
ByronHsu, line 27: what if they want to specify different http/mirrors for different registries?

pkg/app/bootstrap.go
kemingy, line 335: Should this be a method of the `buildkitutil.BuildkitConfig`?

pkg/app/bootstrap.go
```diff
@@ -315,3 +331,29 @@ func buildkit(clicontext *cli.Context) error {
...
+		parts := strings.Split(reg, ",")
+		if len(parts) != 4 {
+			return fmt.Errorf("invalid registry string: %s", reg)
+		}
+
+		config.Registry = append(config.Registry, strings.Split(parts[0], "=")[1])
+		caPath := strings.Split(parts[1], "=")[1]
+		config.Ca = append(config.Ca, caPath)
+		config.Key = append(config.Key, strings.Split(parts[2], "=")[1])
+		config.Cert = append(config.Cert, strings.Split(parts[3], "=")[1])
```
kemingy, line 348: This has an assumption: the order has to be `register,ca,key,cert` but this has not been validated.

BTW, it seems these 3 files have to be in the same directory as the CA path. Otherwise the bind won't work.

ByronHsu (commented):
General concerns
1. I think naming as `registry-credential` makes more sense then `registry` because the main purpose is to pass credential to enable auth with the registry. 
2. I feel the string parsing logic is a bit overwhelming now. Users have to write bunch of stuff with `;` and `,`, which is kinda error-prone. I feel passing the whole thing as a json string would be easier to handle and validate. For example:

```
envd bootstrap --registry-credential='{
  "Registry": ["reg1", "reg2"],
  "Ca": ["ca1, "ca2"],
   ...etc
}'
```

and you can use golang json to unmarshal the string like

```
type MyData struct {
	List []string `json:"list"`
}

func main() {
	app := &cli.App{
		Name:  "myapp",
		Usage: "An example CLI",
		Flags: []cli.Flag{
			&cli.StringFlag{
				Name:  "json",
				Usage: "JSON string argument",
			},
		},
		Action: func(c *cli.Context) error {
			jsonString := c.String("json")
			if jsonString != "" {
				var data MyData
				err := json.Unmarshal([]byte(jsonString), &data)
				if err != nil {
					return err
				}
				// Process the JSON data here
				fmt.Printf("List items: %+v\n", data.List)
			}
			return nil
		},
	}

	err := app.Run(os.Args)
	if err != nil {
		log.Fatal(err)
	}
}
```

3. Should `use-http` and `dockerhub-mirror` also be included in `registry-credential`? i see they are both for per registry

pkg/app/bootstrap.go
ByronHsu: yeah so i think passing a json string might be better

pkg/app/bootstrap.go
```diff
@@ -112,56 +107,70 @@ func bootstrap(clicontext *cli.Context) error {
...
-	mirror := clicontext.String("dockerhub-mirror")
-	if len(mirror) == 0 {
-		return errors.New("`registry-ca-keypair` should be used with `dockerhub-mirror`")
-	}
+	// Loop over all registry strings
+	if clicontext.IsSet("registry") {
+		for _, reg := range strings.Split(clicontext.String("registry"), ";") {
+			// Split into parts
+			parts := strings.Split(reg, ",")
+			if len(parts) != 4 {
```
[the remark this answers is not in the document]
RichhLi: Yes. From our findings, ca is required. Interestingly, cert and key weren't necessary for us, but we wanted to keep the existing structure as generic as possible. We assume the original writers added the cert and key values for a reason.

pkg/app/bootstrap.go
kemingy: This is copied from the original implementation. The main purpose is to make sure these 3 files are in the same directory since we will mount this directory to the buildkitd container.

pkg/driver/docker/docker.go
```diff
@@ -202,22 +198,22 @@ func (c dockerClient) StartBuildkitd(ctx context.Context, tag, name string, bc *
...
 		Image: tag,
 	}
+
 	hostConfig := &container.HostConfig{
 		Privileged: true,
 		AutoRemove: true,
 	}
 
-	if bc.SetCA {
-		hostConfig.Mounts = append(hostConfig.Mounts, mount.Mount{
```
ByronHsu, line 211: Will removing this cause any side effect?

pkg/app/bootstrap.go
ByronHsu: So we basically copy 3 pems to the new dir? If that's the case, do we still need binding to allow the permission to `/etc` or other folders? I think buildkit will have permission to the new dir?

ByronHsu: What do you mean the original impl? and Why not just use `io.Copy` instead of `read` and `write`?

kemingy: > Why not just use `io.Copy` instead of `read` and `write`?

That should be better.

> So we basically copy 3 pems to the new dir? If that's the case, do we still need binding to allow the permission to `/etc` or other folders? I think buildkit will have permission to the new dir?

Need to mount a directory to the buildkitd container. This directory is used to store the envd config, which should be safe.

This PR supports multiple registries which means it needs different CAs for different registries. It might need to differentiate the file names.

pkg/util/buildkitutil/buildkit.go
gaocegege: Maybe we should make the mirror a string slice.

RichhLi: Sure I'll make that change.

gaocegege: cc @kemingy WDYT

pkg/app/bootstrap.go
```diff
@@ -55,32 +56,33 @@ var CommandBootstrap = &cli.Command{
...
-			Name:  "use-http",
-			Usage: "Use HTTP instead of HTTPS for the registry",
-			Value: false,
-		},
 		&cli.StringFlag{
-			Name:    "registry",
-			Usage:   "Specify the registry to pull the image from",
-			Aliases: []string{"r"},
+			Name:  "registry-config",
+			Usage: "Path to a JSON file containing registry configuration",
```
kemingy, line 67: If we're going to pass a JSON file, why not use the buildkit TOML file directly?

I think the main purpose of the CLI is to make it user-friendly. If the situation is complex, it's almost the same as setting up a buildkitd instance manually.

This decision will heavily affect the implementation. What do you think?

cc @gaocegege @RichhLi @ByronHsu

gaocegege: How about providing the buildkit TOML directly to envd?

RichhLi: Very good point. The problem with directly providing a TOML file is that the Docker mount won't be set up. I think this pr may not even be necessary because we might as well just set up buildkitd manually like you said. We will discuss this.

RichhLi: @kemingy @gaocegege Would it be possible to keep the current envd implementation, and the only change we'll make is adding a command that accepts JSON input? And we'll add a check where users can only use one of the commands, not both at the same time. So it will be completely backward compatible.

kemingy: > @kemingy @gaocegege Would it be possible to keep the current envd implementation, and the only change we'll make is adding a command that accepts JSON input? And we'll add a check where users can only use one of the commands, not both at the same time. So it will be completely backward compatible.

I'm okay with it.

gaocegege: LGTM!

Will you implement in this PR?

RichhLi: Yes, or do you want it in different pr?

gaocegege: No, I am just asking. Keeping in this PR LGTM.

gaocegege: Hi @RichhLi Do you still need this PR?

RichhLi: No, I merged this one: https://github.com/tensorchord/envd/pull/1680

kemingy:
close due to https://github.com/tensorchord/envd/pull/1680

PR #1680, opened 2023-07-02
support using json for registry config

RichhLi:
`envd bootstrap --registry-config path_to_json_file`

Format:
Screenshot 2023-07-03 at 3 03 15 PM

Note that the "name" field is required. The other fields are all optional.

This will allow users to specify private Docker registries and the respective credentials. Under the hood, we parse these credentials and create a corresponding buildkit TOML file, as well as set up Docker bindings. This enables users to pull and push images from private Docker registries.

This is also completely backwards compatible with the existing "registry" and "registry-ca-keypair" command. We check that a user only specifies "registy-config" on its own, meaning that a user can't use both registry-config and registry or both registry-config and registry-ca-keypair.

The registry-config command also enables users to specify more than one registry now. For example, if the registry you pull your image from is different than the registry you push your image to, and you need auth for both of them, you should use the registry-config command.

### DIFF
--- a/pkg/app/bootstrap.go
+++ b/pkg/app/bootstrap.go
@@ -76,6 +76,7 @@ var CommandBootstrap = &cli.Command{
 			Name:    "registry",
 			Usage:   "Specify the registry to pull the image from",
 			Aliases: []string{"r"},
+			Value:   "docker.io",
 		},
 		&cli.StringFlag{
 			Name:  "registry-config",
@@ -84,18 +85,6 @@ var CommandBootstrap = &cli.Command{
 	},
 
 	Action: bootstrap,
-}
-
-type Registry struct {
-	Name    string `json:"name"`
-	Ca      string `json:"ca"`
-	Cert    string `json:"cert"`
-	Key     string `json:"key"`
-	UseHttp bool   `json:"use_http"`
-}
-
-type RegistriesData struct {
-	Registries []Registry `json:"registries"`
 }
 
 func bootstrap(clicontext *cli.Context) error {
@@ -136,22 +125,21 @@ func registryCA(clicontext *cli.Context) error {
 	ca := clicontext.String("registry-ca-keypair")
 	registry := clicontext.String("registry")
 
+	if len(ca) == 0 {
+		return nil
+	}
+
 	// We only need this check in registryCA because it is called before registryJSONConfig
 	if len(configFile) > 0 && len(ca) > 0 {
 		return errors.New("only one of `registry-config` and `registry-ca-keypair` can be used")
 	}
-	if len(configFile) > 0 && len(registry) > 0 {
-		return errors.New("only one of `registry-config` and `registry` can be used")
-	}
-	if len(ca) == 0 {
-		return nil
-	}
+
 	mirror := clicontext.String("dockerhub-mirror")
 	if len(mirror) == 0 {
 		return errors.New("`registry-ca-keypair` should be used with `dockerhub-mirror`")
 	}
 
-	// parse ca/key/cert
+	// Parse ca/key/cert
 	kvPairs := strings.Split(ca, ",")
 	if len(kvPairs) != 3 {
 		return errors.New("`registry-ca-keypair` requires ca/key/cert 3 part separated by ','")
@@ -175,10 +163,6 @@ func registryCA(clicontext *cli.Context) error {
 		}
 		if !exist {
 			return errors.Newf("file %s doesn't exist", kv[1])
-		}
-
-		if len(registry) == 0 {
-			registry = "docker.io"
 		}
 
 		path, err := fileutil.ConfigFile(fmt.Sprintf("%s_%s.pem", registry, kv[0]))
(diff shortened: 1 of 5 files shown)


pkg/util/buildkitutil/buildkit.go
```diff
@@ -20,21 +20,33 @@ import (
...
-	Mirror   string
-	UseHTTP  bool
-	SetCA    bool
+	RegistryName []string
+	CaPath       []string
+	CertPath     []string
+	KeyPath      []string
+	UseHTTP      []bool
+	Mirror       string
 }
```
kemingy, line 50: Can we use `RegistriesData` directly? Thus we don't need to do the redundant `append` after the JSON parsing. BTW I think the name `BuildkitConfig` is more suitable.

pkg/app/bootstrap.go
```diff
@@ -146,7 +176,12 @@ func registryCA(clicontext *cli.Context) error {
 		if !exist {
 			return errors.Newf("file %s doesn't exist", kv[1])
 		}
-		path, err := fileutil.ConfigFile(fmt.Sprintf("registry_%s.pem", kv[0]))
+
+		if len(registry) == 0 {
+			registry = "docker.io"
+		}
+
+		path, err := fileutil.ConfigFile(fmt.Sprintf("%s_%s.pem", registry, kv[0]))
```
kemingy, line 184: Will it be better to set `docker.io` as the default value for the argument `registry`?

RichhLi: I set it to `docker.io` because that's what the original config template set it to. Is there a better default? 

In practice however, I assume a user will use both `registry` and `registry-ca-keypair` together, otherwise you're providing credentials to an unknown registry. Maybe we should add a check to ensure they use both?

kemingy: `registry-ca-keypair` should require the `registry`. I mean you can set a default value like https://github.com/tensorchord/envd/blob/5383e01882204da67a68eff856110807b31130f5/pkg/app/bootstrap.go#L51. So you don't need to check if it's an empty string.

pkg/util/buildkitutil/buildkit.go
RichhLi: Sure that's possible. But I think then we would have to change the JSON format. Something like this:

`{
    "mirror": ["dockerhub-mirror", "dockerhub-mirror"]
    "registry_name": ["myRegistry1", "myRegistry2"],
    "ca_path": ["myCa1", "myCa2"],
    "key_path": ["", "myKey2"],
    "cert_path": ["", "myCert2"],
    "use_http": [false, false]
}`

The format I proposed in the pr is nested, so I don't think unmarshaling directly into BuildkitConfig is possible. But if we change the JSON format to the above, I think it's possible.

What do you think? Personally I think its a tradeoff between user readability and code simplicity. But if we're being practical, I don't think a user will be specifying more than two registries, so I think the decrease in readability is ok.

kemingy: I think we don't have to change the JSON format. The current format used by `RegistriesData` (might need to add the mirror to `Registry`) LGTM. I mean they can be the same struct. No need to maintain two similar structs.

RichhLi: To clarify, are you suggesting to move `RegistriesData` and `Registry` to `buildkit.go`? And then we would have to change `buildkitConfigTemplate` right? Because right now `buildkitConfigTemplate` is based on the format of `BuildkitConfig`.

kemingy: Yes.

gaocegege:
Hi is it ready for another review?

RichhLi:
Yes sorry, I'm new to Github and open source. In general if I make a change, how do I notify reviewers? Just leave a comment?

kemingy:
> Yes sorry, I'm new to Github and open source. In general if I make a change, how do I notify reviewers? Just leave a comment?

You can click the "re-request review" on the top right reviewers' list. Leaving a comment is also okay.

pkg/app/bootstrap.go
```diff
@@ -289,14 +371,33 @@ func buildkit(clicontext *cli.Context) error {
...
+			Name:    clicontext.String("registry"),
+			Ca:      "/etc/registry",
+			Cert:    "/etc/registry",
+			Key:     "/etc/registry",
+			UseHttp: clicontext.Bool("use-http"),
+			Mirror:  clicontext.String("dockerhub-mirror"),
+		})
+		// The values of Ca, Cert, and Key don't actually matter since we already copied their contents to /etc/registry
+		// So instead of parsing registry-ca-keypair again, we'll just put an arbitrary value
+		// This is to ensure that buildkitConfigTemplate parses properly
```
kemingy, line 397:
```suggestion
		// The values of Ca, Cert, and Key don't actually matter since we already copied their contents to the envd config directory and mounted to `/etc/registry`.
		// So instead of parsing registry-ca-keypair again, we'll just put the default value.
		// This is to ensure that buildkitConfigTemplate parses properly.
		config.Registries = append(config.Registries, buildkitutil.Registry{
			Name:    clicontext.String("registry"),
			Ca:      "/etc/registry",
			Cert:    "/etc/registry",
			Key:     "/etc/registry",
			UseHttp: clicontext.Bool("use-http"),
			Mirror:  clicontext.String("dockerhub-mirror"),
		})
```

kemingy (approved):

PR #1699, opened 2023-07-10
feat(v1): disable diff/merge op for moby builder

### LINKED ISSUE #1693: feat: disable buildkit merge op when using moby builder
### Describe the feature

Currently, we use `docker` instead of `moby` builder because https://github.com/tensorchord/envd/issues/1547.

- track the upstream issue https://github.com/moby/moby/issues/45111

The downside is that we need to convert the image type which may take a long time for large CUDA-based images.

A possible solution is to disable the merge op when using moby builder.

Currently, merge op is used by:
- system package & language installation
- extra HTTP sources (rare)
- language packages (when the build has multiple languages, i.e. `python` + `r`) (rare)

### Why do you need this feature?

_No response_

### Additional context

_No response_

kemingy:
- fix #1693

### DIFF
--- a/pkg/lang/ir/v1/compile.go
+++ b/pkg/lang/ir/v1/compile.go
@@ -29,6 +29,7 @@ import (
 	servertypes "github.com/tensorchord/envd-server/api/types"
 
 	"github.com/tensorchord/envd/pkg/config"
+	"github.com/tensorchord/envd/pkg/home"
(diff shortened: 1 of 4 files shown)


pingsutw:
@kemingy thanks, let me test it first, I'll get back to you.

gaocegege (approved):

PR #1708, opened 2023-07-17
feat(v1): support moby pushing image

kemingy:
The default progress bar (guess it's copied from buildkit) is tricky to use. So I switch to another one. It might be slightly different.


gaocegege (approved):

pingsutw:
I just tested it, and it works for me, thanks

Issue #1519, opened 2023-03-11
feat: support gpu more than one


aseaday:
https://github.com/tensorchord/envd/blob/cb90327362f768ba37b3d03b1524dd3535156acf/pkg/app/up.go#L218

https://github.com/tensorchord/envd/blob/cb90327362f768ba37b3d03b1524dd3535156acf/pkg/app/up.go#L200

It seems that we only support GPU=1, more and more developers use a multi card server to develop model. Maybe we should give a settings here.

aseaday:
It could be configure in config.gpu.

aseaday:
It doesn't work config.gpu.

aseaday:
I have two ideas to fix this problems:
1. add an option for `envd up` and remove `config.gpu`
2. make `config.gpu` work.
I prefer the first solution. But maybe I need more information of envd server's design purpose.

VoVAllen:
I think we can make it a cli args for now. Config.gpu will only add a label to the image. The runtime behavior depends on the actual runner

aseaday:
I will add a cli arg for control gpus.

arugal:
@aseaday Hi, this feature is very useful to me. Is there any progress on it now? Maybe I can help.

aseaday:
I will push my commit asap.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #1730, opened 2023-08-13
feat: envd up with `--gpus`

kemingy:
- close https://github.com/tensorchord/envd/issues/1519


kemingy:
This is slightly different from `docker run --gpus` since we only support indicating the required number. Let me know if it's necessary to support a more complex input. @gaocegege @aseaday @arugal

gaocegege (approved):

PR #1766, opened 2023-09-11
feat: add the reference subcommand to generate the markdown doc

kemingy:
We can maintain the reference code in this repo. Will delete the go module and go code in `envd-docs`.


kemingy:
cc @VoVAllen @gaocegege PTAL

PR #1824, opened 2023-12-28
fix: update docker/docker dependency

kemingy:
- fix the release error caused by `go mod tidy`: https://github.com/tensorchord/envd/actions/runs/7343727313/job/19994652738


kemingy:
CI failed because of the previous release failure. Will merge it directly.

PR #1843, opened 2024-01-24
fix: allow cli --gpus to override config gpu

kemingy:
Previously, `config.gpu()` will be ignored in `envd up`.


iodone:
Very efficient bug handling, awesomes!

PR #1863, opened 2024-03-11
chore: use go stable version

kemingy:
Previously, we had to pin the go version to avoid the HTTP Host panic. Now, it's already fixed. Let's change to the stable version.

- refer to https://github.com/tensorchord/envd/pull/1713


kemingy:
PTAL @gaocegege

gaocegege (approved):

PR #1939, opened 2025-01-08
feat: upgrade to ubuntu 22.04

kemingy:
Since Ubuntu 20.04 LTS will reach the EOL this April, we should upgrade to 22.04 now.

- https://ubuntu.com/about/release-cycle

cc @gaocegege WDYT


.github/workflows/release.yml
```diff
@@ -47,7 +47,7 @@ jobs:
     timeout-minutes: 20
     strategy:
       matrix:
-        os: [macos-13, ubuntu-20.04]
+        os: [macos-11, ubuntu-22.04]
```
gaocegege, line 50: It should be 13.

gaocegege:
Generally LGTM!

.github/workflows/release.yml
kemingy: Fixed.

gaocegege (approved):

Issue #1974, opened 2025-02-07
bug: cannot export cache with moby builder


kemingy:
### Are you use the envd server?

- [ ] Yes, I am using the envd server.
- [x] No, I am not using the envd server.

### Describe the bug

- release failed on cache-export: https://github.com/tensorchord/envd/actions/runs/13192340600/job/36827652885

### To Reproduce

- use envd:v1.0.0-alpha.2

```sh
envd build -f base-images/remote-cache/build.envd:build --export-cache type=registry,ref=docker.io/kem
ingy/python-cache:envd-v1.0.0-alpha.2
```

### The `docker info` output

```
Client: Docker Engine - Community
 Version:           27.5.1
 API version:       1.47
 Go version:        go1.22.11
 Git commit:        9f9e405
 Built:             Wed Jan 22 13:42:33 2025
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          27.5.1
  API version:      1.47 (minimum version 1.24)
  Go version:       go1.22.11
  Git commit:       4c9b3b0
  Built:            Wed Jan 22 13:40:53 2025
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.7.25
  GitCommit:        bcc810d6b9066471b0b6fa75f557a15a1cbf31bb
 runc:
  Version:          1.2.4
  GitCommit:        v1.2.4-0-g6c52b3f
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### The `envd version` output

```
envd: v1.0.0-alpha.2+2b3a8d1.dirty
  BuildDate: 2025-02-07T03:58:54Z
  GitCommit: 2b3a8d184acaaf286c56ada79cd2e3c5bd646c6c
  GitTreeState: dirty
  GitTag: v1.0.0-alpha.2
  GoVersion: go1.23.4
  Compiler: gc
  Platform: linux/amd64
```

kemingy:
The `moby/buildkit:v0.19.0` works as expected.

kemingy:
It turns out the default docker driver can only support those cache backends when the `containerd` image store is enabled.

- https://docs.docker.com/build/cache/backends/
- https://docs.docker.com/engine/storage/containerd/

So, the easiest way is to use a `moby/buildkit` container. 😢 

I'll document this in the README.

PR #1975, opened 2025-02-07
fix: use buildkit container builder to export cache

kemingy:
- fix https://github.com/tensorchord/envd/issues/1974


gaocegege (approved):

gaocegege:
You wake up so early

PR #1992, opened 2025-03-09
chore: bump dep version, check dep monthly

kemingy:


pkg/lang/frontend/starlark/v1/interpreter.go
```diff
@@ -42,6 +43,19 @@ type entry struct {
...
+	return &syntax.FileOptions{
+		// resolver
+		Set:               false,
+		While:             false,
+		TopLevelControl:   false,
+		GlobalReassign:    false,
+		LoadBindsGlobally: false,
+		// compiler
+		Recursion: true,
+	}
```
kemingy, line 56: The latest version of starlark requires explicit options instead of legacy global options.

The configurations here should work for most of the cases.

gaocegege (approved):

PR #1129, opened 2022-11-01
fix: Fix zsh tab issue

gaocegege:
Close #183 

Test case:

```
def build():
    shell("zsh")
```

Before the PR:

```
~/golang via 🅒 envd 
⬢ [envd]❯ l<tab>

~/golang via 🅒 envd 
⬢ [envd]❯ l   l
l              ldconfig.real  LISTMAX        logger         lsblk          lzcmp        
la             ldd            ll             login          LS_COLORS      lzdiff       
langinfo       LESS           ln             logname        LSCOLORS       lzegrep      
last           let            local          LOGNAME        lscpu          lzfgrep   
```

After the PR:

```
~/golang via 🅒 envd 
⬢ [envd]❯ l<tab>
~/golang via 🅒 envd 
⬢ [envd]❯ l
l              ldconfig.real  LISTMAX        logger         lsblk          lzcmp        
la             ldd            ll             login          LS_COLORS      lzdiff       
langinfo       LESS           ln             logname        LSCOLORS       lzegrep      
last           let            local          LOGNAME        lscpu          lzfgrep   
```

Signed-off-by: Ce Gao <cegao@tensorchord.ai>

PR #2010, opened 2025-04-01
fix: locale for both dev=True/False

kemingy:
- revert https://github.com/tensorchord/envd/pull/1129, we use `starship` for prompt
- update miniconda
- related to https://github.com/tensorchord/envd/discussions/2009


kemingy:
> The difference between C. and en_US. primarily affects formatting rules (like dates, numbers, currency) and collation (sorting order). It does not affect the ability to encode or decode specific characters, as long as the encoding is declared as UTF-8.

gaocegege (approved):

PR #2040, opened 2025-10-01
chore: update the code owners

kemingy:
This project has been in maintenance status for a long time. Some small updates (to fix the breaking changes introduced by bumping the dependency version) automatically invite several people to review, which is usually unnecessary.

We really appreciate their past contributions. 

**This doesn't change any permissions, only for the automatically generated PR review invitation.** 

We'll still manually invite them to review the code when it's necessary.


kemingy:
cc @gaocegege

